Issue #7, opened 2015-06-04
Make an ACL engine like the Authentication


devinsba:
Allow ACL to pluggable like the authentication currently is.

Ex:
* LDAP
* A database

This allows an organization to have non developer maintain this mapping. For instance with new hires or people leaving. Or new repos and new groups

I would be happy to take a shot at adding this if you are open to this type of feature. Certain implementations would also solve #6

rojer:
yes, this sounds reasonable to me.

rojer:
fixed by #23

PR #16, opened 2015-06-25
Factor out authenticators into a separate package

rojer:
Just shuffling stuff around, no real changes

#7

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/16)


gelraen:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[auth_server/authn/authn.go, line 27 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/16#-JseB3aE3ICQNzkD7Qvy)** ([raw file](https://github.com/cesanta/docker_auth/blob/7c7ad9ed8f3d4e7a3e6bc932959df6fd587b7489/auth_server/authn/authn.go#L27)):</sup>
This is a pretty strict requirement for the users of the interface, which forces them to have synchronization, even though implementation should be goroutine-safe as well. Additional reasoning why it is done this way would reduce the confusion.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/16)

rojer:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[auth_server/authn/authn.go, line 27 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/16#-JseB3aE3ICQNzkD7Qvy)** ([raw file](https://github.com/cesanta/docker_auth/blob/7c7ad9ed8f3d4e7a3e6bc932959df6fd587b7489/auth_server/authn/authn.go#L27)):</sup>
it's actually not a requirement but a guarantee, which makes it easier for implementors :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/16)

gelraen:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[auth_server/authn/authn.go, line 27 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/16#-JseB3aE3ICQNzkD7Qvy)** ([raw file](https://github.com/cesanta/docker_auth/blob/7c7ad9ed8f3d4e7a3e6bc932959df6fd587b7489/auth_server/authn/authn.go#L27)):</sup>
Not that much, they still have to take care to make the implementation goroutine-safe. And the user must also take care (independetly of the implementation) to not call Authenticate after Stop and not call Stop until all Authenticate calls return.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/16)

rojer:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[auth_server/authn/authn.go, line 27 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/16#-JseB3aE3ICQNzkD7Qvy)** ([raw file](https://github.com/cesanta/docker_auth/blob/7c7ad9ed8f3d4e7a3e6bc932959df6fd587b7489/auth_server/authn/authn.go#L27)):</sup>
to the goroutine-safe requirement my response is - tough cookie :)
(and i just realized taht with the addition of leveldb, my google auth is actually not safe, so i'll fix that).
but the guarantee provided for Stop actually makes synchronization easier. implementations may or may not want to make use of it, but it's there.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/16)

PR #23, opened 2015-08-20
Auth{n,z} refactor; fix token generation for 1.8

### LINKED ISSUE #22: Can no longer pull with docker 1.8.1 if the user only has pull access
see

https://github.com/docker/docker/issues/15640

I guess it's supposed to return the token with the appropriate rights and it's the registries responsibility to deny access.

Which makes since from an SSO kind of implementation with a JWT.

in 1.8.1 docker is requesting push and pull for just a pull, not sure why they changed that.

rojer:
Since 1.8 Docker will always request push and pull actions (scopes),
and we should return the subset that is allowed.

Fixes #22
docker/docker#15640

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/23)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ See the [example config files](https://github.com/cesanta/docker_auth/tree/maste
 
 Run with increased verbosity:
 ```{r, engine='bash', count_lines}
-docker run ... cesanta/docker_auth --v=2 /config/auth_config.yml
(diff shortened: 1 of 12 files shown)


gelraen:
Reviewed 12 of 12 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/23)

PR #25, opened 2015-08-28
Allow non tls server

dkjer:
This PR allows running the auth server as http:// instead of https://.  Our use-case is that we prefer to configure https at the ELB level, not at the internal node level.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/25)

### DIFF
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -46,20 +46,29 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
 		glog.Exitf("Failed to create auth server: %s", err)
 	}
 
-	hs := &http.Server{
-		Addr:    c.Server.ListenAddress,
-		Handler: as,
-		TLSConfig: &tls.Config{
+	var tlsConfig *tls.Config
+	if c.Server.CertFile != "" || c.Server.KeyFile != "" {
+		// Check for partial configuration.
+		if c.Server.CertFile == "" || c.Server.KeyFile == "" {
+			glog.Exitf("Failed to load certificate and key: both were not provided")
+		}
+		tlsConfig = &tls.Config{
 			NextProtos:   []string{"http/1.1"},
 			Certificates: make([]tls.Certificate, 1),
-		},
+		}
+		glog.Infof("Cert file: %s", c.Server.CertFile)
+		glog.Infof("Key file : %s", c.Server.KeyFile)
+		tlsConfig.Certificates[0], err = tls.LoadX509KeyPair(c.Server.CertFile, c.Server.KeyFile)
+		if err != nil {
+			glog.Exitf("Failed to load certificate and key: %s", err)
+		}
+	} else {
+		glog.Warning("Running without TLS")
 	}
-
-	glog.Infof("Cert file: %s", c.Server.CertFile)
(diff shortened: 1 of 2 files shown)


rojer:
i considered doing that, but then thought that since token signing cert is required anyway, it won't be of any benefit. hence the ability to use one cert for both signing and TLS.
is there any reason why that doesn't work for you?

rojer:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[auth_server/main.go, line 50 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/25#-JxrQRRMUGJxEGVTnE95)** ([raw file](https://github.com/cesanta/docker_auth/blob/ad4b3fda8fec6f8090a1dedf52b1d2d6bdef1242/auth_server/main.go#L50)):</sup>
same here: have it load the the cert/key if any of them is specified, which will fail if one isn't, but that's fine.

---

<sup>**[auth_server/server/config.go, line 125 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/25#-JxrQMNIpYPYvQ33OtLX)** ([raw file](https://github.com/cesanta/docker_auth/blob/ad4b3fda8fec6f8090a1dedf52b1d2d6bdef1242/auth_server/server/config.go#L125)):</sup>
make a case where only one is specified an error, as it doesn't make sense to specify only cert or only key.

---

<sup>**[auth_server/server/config.go, line 138 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/25#-JxrPVg6Oxm4xy9VCZce)** ([raw file](https://github.com/cesanta/docker_auth/blob/ad4b3fda8fec6f8090a1dedf52b1d2d6bdef1242/auth_server/server/config.go#L138)):</sup>
actually, both are required

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/25)

dkjer:
@rojer Using one cert for both signing and TLS does work, and we may end up doing that for production.  However we prefer to decouple that for now, at least in our dev environment.

rojer:
@dkjer: still not sure i understand the reasoning, but will merge once comments are addressed.

dkjer:
Rebased with comments addressed

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/25)

rojer:
Reviewed 2 of 2 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/25)

rojer:
built and pushed to :latest

dkjer:
Thanks @rojer !

djenriquez:
@rojer, @dkjer I see that this has been merged, but how does one set up the registry config to use this? The registry config hard requries a rootcertbundle, but if TLS is disabled?

I'm also looking at doing TLS termination on the ELB rather than in the auth server.

Could you provide examples of both the registry config and the auth server config?

Thanks!!!

PR #27, opened 2015-09-02
support ldap authentication

summerQLin:
Signed-off-by: Lin Qing <qing.lin@hp.com>

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/27)

### DIFF
--- a/auth_server/Dockerfile
+++ b/auth_server/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.4
+FROM golang:1.5
 EXPOSE 5001
 ADD auth_server /
 ENTRYPOINT ["/auth_server"]
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -1,0 +1,146 @@
+/*
+   Copyright 2015 Cesanta Software Ltd.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       https://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package authn
+
+import (
+	"bytes"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/go-ldap/ldap"
+	"github.com/golang/glog"
+)
+
+type LdapAuthConfig struct {
+	Domain          string   `yaml:"domain,omitempty"`
+	Port            uint16   `yaml:"port,omitempty"`
+	StartTLS        bool     `yaml:"startTLS,omitempty"`
+	BaseDN          string   `yaml:"baseDN,omitempty"`
+	LoginAttributes []string `yaml:"loginAttribute,omitempty"`
+	GroupBaseDN     string   `yaml:"groupBaseDN,omitempty"`
+	GroupAttribute  string   `yaml:"groupAttribute,omitempty"`
+}
+
+type LdapAuth struct {
+	config *LdapAuthConfig
+}
+
+func NewLdapAuth(c *LdapAuthConfig) (*LdapAuth, error) {
+	return &LdapAuth{
+		config: c,
+	}, nil
+}
+
+func (la *LdapAuth) Authenticate(user string, password PasswordString) (bool, error) {
+	if user == "" {
+		return true, nil
+	}
+	l, err := la.ldapConnection()
+	if err != nil {
+		return false, err
+	}
+	defer l.Close()
+	//l.Debug = true
+	filter := la.getLoginFilter(user)
+	userEntryDN, uSearchErr := la.ldapSearch(l, &la.config.BaseDN, &filter, &[]string{})
+	if uSearchErr != nil {
+		return false, uSearchErr
+	}
+	if len(userEntryDN) > 0 {
+		err := l.Bind(userEntryDN, string(password))
+		if err != nil {
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+func (la *LdapAuth) Name() string {
+	return "Ldap"
+}
+
+func (la *LdapAuth) ldapConnection() (*ldap.Conn, error) {
+	glog.V(2).Infof("Dial: starting...%s:%d", la.config.Domain, la.config.Port)
+	l, err := ldap.Dial("tcp", fmt.Sprintf("%s:%d", la.config.Domain, la.config.Port))
+	if err != nil {
+		return nil, err
+	}
+	if la.config.StartTLS {
+		glog.V(2).Infof("StartTLS...")
+		err = l.StartTLS(&tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return l, nil
+}
+
+//make filter by login attributes, e.g. login in by ['cn', 'uid']
+//the filter will be '(|(cn=account)(uid=account))'
+func (la *LdapAuth) getLoginFilter(user string) string {
+	var buffer bytes.Buffer
+	buffer.WriteString("(|")
+	for _, attr := range la.config.LoginAttributes {
+		buffer.WriteString(fmt.Sprintf("(%s=%s)", attr, user))
+	}
+	buffer.WriteString(")")
+	return fmt.Sprintf(buffer.String())
+}
+
+//ldap search and return required attributes' value from searched entries
+//default return entry's DN value if you leave attrs array empty
+func (la *LdapAuth) ldapSearch(l *ldap.Conn, baseDN *string, filter *string, attrs *[]string) (string, error) {
+	if l == nil {
+		return "", errors.New("No ldap connection!")
+	}
+	glog.V(2).Infof("Searching...basedDN:%s, filter:%s", *baseDN, *filter)
+	searchRequest := ldap.NewSearchRequest(
+		*baseDN,
+		ldap.ScopeWholeSubtree, ldap.NeverDerefAliases, 0, 0, false,
+		*filter,
+		*attrs,
+		nil)
+	sr, err := l.Search(searchRequest)
+	if err != nil {
+		return "", err
+	}
+
+	if len(sr.Entries) != 1 {
+		return "", errors.New("Error...Search Result number != 1\n")
+	}
+
+	var buffer bytes.Buffer
+	for _, entry := range sr.Entries {
+		if len(*attrs) == 0 {
+			glog.V(2).Infof("Entry DN = %s", entry.DN)
+			buffer.WriteString(entry.DN)
+		} else {
+			for _, attr := range *attrs {
+				values := strings.Join(entry.GetAttributeValues(attr), " ")
+				glog.V(2).Infof("Entry %s = %s", attr, values)
+				buffer.WriteString(values)
+			}
+		}
+	}
+
+	return buffer.String(), nil
+}
+
+func (la *LdapAuth) Stop() {
+}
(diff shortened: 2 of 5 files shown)


summerQLin:
Hi rojer,

I add a ldap authenticator. To see how this works, you can check examples/ldap_config.yml. 
I am looking forward to your feedback.

rojer:
thank you! this looks ok overall, but i made a few comments. once addressed, this should be good to go.

---

Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

<sup>**[auth_server/authn/ldap_auth.go, line 30 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCJ4C9-nRiX4cWsPP8)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L30)):</sup>
s/Ldap/LDAP/ everywhere

---

<sup>**[auth_server/authn/ldap_auth.go, line 31 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCNsqFRWbTVqq6Luyf)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L31)):</sup>
you join domain and port anyway, use a single "addr" config option instead

---

<sup>**[auth_server/authn/ldap_auth.go, line 33 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCJPmsffHF4fEP2tYn)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L33)):</sup>
lower_case_and_underscores instead of camelCase in YAML attribute names, please. hence: start_tls, base_dn, login_attributes, group_base_dn, group_attribute

---

<sup>**[auth_server/authn/ldap_auth.go, line 51 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCKrh_3QgmnrJRQTeL)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L51)):</sup>
this implicitly allows anonymous access. authenticators shouldn't do this. instead, anonymous access, if desired, shoudl be explicitly enabled through static auth by adding this to the config:

users:
  "": {}  # Allow anonymous (no "docker login") access.

instead, return false, NoMatch here

---

<sup>**[auth_server/authn/ldap_auth.go, line 59 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCJsfIo213elFxMQ1k)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L59)):</sup>
remove

---

<sup>**[auth_server/authn/ldap_auth.go, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCLwaOJNVRDizKgfbX)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L100)):</sup>
this is susceptible to "Bobby Tables attack" (https://xkcd.com/327/) :)
username is supplied by the user and you need to make sure it's properly quoted for inclusion in LDAP query

---

<sup>**[auth_server/authn/ldap_auth.go, line 125 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCMPjrR06H_g5GVKS5)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/authn/ldap_auth.go#L125)):</sup>
remove \n

---

<sup>**[auth_server/server/server.go, line 51 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCIeUinkH9E6kPrd89)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/server/server.go#L51)):</sup>
since ldap auth doesn't need web handler, there's no need to have this extra variable. adding to the authenticators slice is enough.

---

<sup>**[auth_server/server/server.go, line 131 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCIX0rxe5ApKydqUKe)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/auth_server/server/server.go#L131)):</sup>
remove this extra line

---

<sup>**[examples/ldap_config.yml, line 8 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCNNC-VHWqlOcwWwYn)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/examples/ldap_config.yml#L8)):</sup>
please provide a description of how the authentication works: retrieving user DN by search and trying to bind with the specified password. this is not immediately obvious.

---

<sup>**[examples/ldap_config.yml, line 13 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyCMivaX7hPZZm0VO7q)** ([raw file](https://github.com/cesanta/docker_auth/blob/d7d09183c1cb7cdb97bd4f1a04d5721cfa8fe0f5/examples/ldap_config.yml#L13)):</sup>
so this is or'ed together? actually, i can easily see someone needing a more complex query. it should be easy to accommodate that by providing a template where ${account} is substituted (with appropriate quoting). we recently did this for authz, see use of strings.Replacer in authz/acl.go.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/27)

summerQLin:
Hi Rojer, 

I changed design of ldap configuration file. Instead of concatenating strings as the filter, the filter template is supported, but only accept ${account} as the variable.

Another issue you addressed is the risk of ldap injection. I am trying to escape some characters, but "\" doesn't work for escaping purpose. I will need more time to check.

But anyway, please help review.

rojer:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

<sup>**[auth_server/authn/ldap_auth.go, line 22 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoL2jEm5l6nIfJWTB4)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/auth_server/authn/ldap_auth.go#L22)):</sup>
what's wrong with just "base64"?

---

<sup>**[auth_server/authn/ldap_auth.go, line 32 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoLCy15t-KSQpq3UVS)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/auth_server/authn/ldap_auth.go#L32)):</sup>
sorry for being nitpicky - let's rename this to just "tls", hence "tls: true/false". the fact that it is implemented via upgrade ("starttls" command or whatever) is an implementation detail.

---

<sup>**[auth_server/authn/ldap_auth.go, line 116 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoTnJt0uJlbirbZZrY)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/auth_server/authn/ldap_auth.go#L116)):</sup>
you'll have to study ways of quoting ${account} to avoid query injection.
i can't help you here, as i'm only superficially familiar with LDAP... you say escaping `"` is not enough? maybe just removing anything that does not match, say, `[a-zA-Z0-9_-]`? that subset should be safe in pretty much any query.

---

<sup>**[examples/ldap_config.yml, line 9 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoMJhyGXKlmO7G77tJ)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/examples/ldap_config.yml#L9)):</sup>
add a comment here that this allows anonymous access and is not necessary for LDAP auth as such.
come to think of it, just remove it altogether - otherwise cargo-culting will have anonymous access enabled in all ldap authenticated registries around the world :)

---

<sup>**[examples/ldap_config.yml, line 11 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoMqgaUod26y7s-KO6)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/examples/ldap_config.yml#L11)):</sup>
looks like 16.217.2.55 is a server in HP. ldap.example.com:389 will do better here :)

---

<sup>**[examples/ldap_config.yml, line 15 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoNEO4Q8-7GuXvmeYy)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/examples/ldap_config.yml#L15)):</sup>
base64? please say it's not for security :)
i know docker uses it in its config files, but this is not something i'd endorse :)
there should be no plain-text credentials in the config files. BCrypt hashes are tolerable because they are one-way and non-reversible. config files are supposed to live in source control, and checking plain text passwords into source control is a no-no.
one alternative is to have a `bind_password_file` parameter, which will specify an external file containing plain text password, presumably not checked into SCM and managed through different means.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/27)

summerQLin:
Hi rojer,
Please check my latest commit. I just found go-ldap/ldap recently has an update, and is able to recognize hex characters in form of \5c (literal meaning is \). Therefore I can escape special characters by converting them to hex.
By the way, the anonymous users here will be controlled by ACL. For our scenario, we allow anonymous users to pull image from a publish path name. Can't the design be like that?

rojer:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

<sup>**[auth_server/authn/ldap_auth.go, line 113 \[r4\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JzVF0XoYn6RFv3HlaOv)** ([raw file](https://github.com/cesanta/docker_auth/blob/c895a72f10d0f8a3e28d9bb014f33b7dfdb6c032/auth_server/authn/ldap_auth.go#L113)):</sup>
string replacer is cool, you can specify all the replacements upfront: 
r = strings.NewReplacer(
    `\`, `\5c`, 
    `(`, `\28`,
...
)

return r.Replace(account)

---

<sup>**[examples/ldap_config.yml, line 9 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JyoMJhyGXKlmO7G77tJ)** ([raw file](https://github.com/cesanta/docker_auth/blob/9a462a19b6653b9b25d262fb66d3310122949a98/examples/ldap_config.yml#L9)):</sup>
please add a comment here that this allows anonymous access

---

<sup>**[examples/ldap_config.yml, line 14 \[r4\]](https://reviewable.io:443/reviews/cesanta/docker_auth/27#-JzVFwqzAW7DcnPMv11i)** ([raw file](https://github.com/cesanta/docker_auth/blob/c895a72f10d0f8a3e28d9bb014f33b7dfdb6c032/examples/ldap_config.yml#L14)):</sup>
comment:

# ${account} will be taken from request.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/27)

rojer:
almost there. just a couple minor nitpicks.
also please rebase on current master, as i've already updated to go 1.5

summerQLin:
Hi rojer,

Please check new commits. 

I removed anonymous access to avoid confusion. Besides, I think I've already updated Dockerfile to use go 1.5, do you mean that?

rojer:
nice. thank you!

PR #28, opened 2015-09-04
Disable broken SSLv3 and force strong ciphers to match Docker Registry 2.1 security

jmastron:
SSLv3 is broken but is still enabled in crypto/tls.  Docker Registry 2.1 has already implemented a workaround, so we should follow their lead here.  See https://github.com/docker/distribution/commit/dc5869de0b4fcab4f0291c8e835560951c5e968a

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/28)


rojer:
thank you!

Issue #37, opened 2015-10-26
Invalid Memory Address with AD LDAP via TLS


kidhasmoxy:
I've configured LDAP for TLS connecting to an AD server, but I get the following error whenever a user attempts to login.

runtime error: invalid memory address or nil pointer dereference
goroutine 12 [running]:
net/http.(*conn).serve.func1(0xc82017fce0, 0x7f2f7223a268, 0xc82008c840)
	/usr/local/go/src/net/http/server.go:1287 +0xb5
github.com/go-ldap/ldap.(*Conn).StartTLS(0xc82027c900, 0xc8201c1680, 0x0, 0x0)
	/home/rojer/go/src/github.com/go-ldap/ldap/conn.go:180 +0x125c
github.com/cesanta/docker_auth/auth_server/authn.(*LDAPAuth).ldapConnection(0xc8200240b0, 0xc8201fe540, 0x0, 0x0)
	/home/rojer/go/src/github.com/cesanta/docker_auth/auth_server/authn/ldap_auth.go:133 +0x33d
github.com/cesanta/docker_auth/auth_server/authn.(*LDAPAuth).Authenticate(0xc8200240b0, 0xc8202748d2, 0xd, 0xc82027e32e, 0x9, 0x0, 0x0, 0x0)
	/home/rojer/go/src/github.com/cesanta/docker_auth/auth_server/authn/ldap_auth.go:56 +0x8c
github.com/cesanta/docker_auth/auth_server/server.(*AuthServer).Authenticate(0xc820126680, 0xc82025ed80, 0x11, 0x0, 0x0)
	/home/rojer/go/src/github.com/cesanta/docker_auth/auth_server/server/server.go:109 +0x11e
github.com/cesanta/docker_auth/auth_server/server.(*AuthServer).doAuth(0xc820126680, 0x7f2f7223e8e8, 0xc8202820b0, 0xc820233960)
	/home/rojer/go/src/github.com/cesanta/docker_auth/auth_server/server/server.go:229 +0x3d1
github.com/cesanta/docker_auth/auth_server/server.(*AuthServer).ServeHTTP(0xc820126680, 0x7f2f7223e8e8, 0xc8202820b0, 0xc820233960)
	/home/rojer/go/src/github.com/cesanta/docker_auth/auth_server/server/server.go:201 +0x1ef
net/http.serverHandler.ServeHTTP(0xc8201a05a0, 0x7f2f7223e8e8, 0xc8202820b0, 0xc820233960)
[...]

PR #40, opened 2015-11-02
enable ldap tls connection

summerQLin:
Fix for issue #37 to enable SSL/TLS connection to LDAP server. Now "plain", "simple_tls", "starttls" is supported. Usually "plain" and "starttls" goes from port 389, and "simple_tls" uses 636.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/40)

### DIFF
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -29,7 +29,7 @@ import (
 
 type LDAPAuthConfig struct {
 	Addr             string `yaml:"addr,omitempty"`
-	Method           string `yaml:"method,omitempty"`
+	TLS              string `yaml:"tls,omitempty"`
 	Base             string `yaml:"base,omitempty"`
 	Filter           string `yaml:"filter,omitempty"`
 	BindDN           string `yaml:"bind_dn,omitempty"`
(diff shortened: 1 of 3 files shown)


rojer:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[auth_server/authn/ldap_auth.go, line 32 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/40#-K280l-ZXsrMdYSkaaLx)** ([raw file](https://github.com/cesanta/docker_auth/blob/00a0f2678bdd12dc83c218444428b582fd902a7b/auth_server/authn/ldap_auth.go#L32)):</sup>
let's keep the param name "tls" but change it to string type and set possible values to "" or "none" (no TLS, the default), "always" (DialTLS), or "starttls".

also, do i understand it correctly that port 686 pretty requires DialTLS (i.e. tls = always in our new lingo)? if so, can we add logic that will set tls = always if config.Addr ends with ":686", unless tls is explicitly specified?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/40)

summerQLin:
Please review.

rojer:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[auth_server/server/server.go, line 129 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/40#-K2LxI_4v3tnIbyQiCBS)** ([raw file](https://github.com/cesanta/docker_auth/blob/a3cdedccc49b81c569ac35060c405c0212965311/auth_server/server/server.go#L129)):</sup>
remove this extra line


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/40)

rojer:
getting there. please add the bit about using tls=always when it's not set (`la.config.TLS == ""`) and ports is 686 (`la.config.Addr.HasSuffix(":686")`)

also, i didn't notice it previously, but you disable cert verification (`InsecureSkipVerify: true`). this renders TLS completely ineffective, enabling MITM attacks by default. this is bad, we don't want this. please add a separate option, `insecure_tls_skip_verify`, that will control this option and make it default to false.

rojer:
thanks, this became https://github.com/cesanta/docker_auth/commit/a9c3c39ec8846cf6d5723a4c37f5372a4a02de13

PR #57, opened 2015-12-17
Add optional sort_keys field for ACLs.

carsonoid:
Without this you are often forced to delete/recreate records to manipulate natural ordering which is not feasible in large deployments.

Example results from my dataset:

Given: ``sort_keys: ["priority", "Comment"]``

Natural order:
```golang
[
    {"Match":{"account":"admin"},"Actions":["*"],"Comment":"Priority 2"}
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"Priority 1"}
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"Apriority 1"}
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"priority none"}
    {"Match":{"account":"admin"},"Actions":["*"],"Comment":"APriority 0"}
]
```

Sorted order:
```golang
[
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"priority none"} 
    {"Match":{"account":"admin"},"Actions":["*"],"Comment":"APriority 0"} 
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"Apriority 1"} 
    {"Match":{"account":"/.+/","name":"${account}/*"},"Actions":["*"],"Comment":"Priority 1"} 
    {"Match":{"account":"admin"},"Actions":["*"],"Comment":"Priority 2"}
]
```

**Note** that mongo will continue to work as it used to without sort_keys defined so this won't change anything for existing users. It's a fully-optional feature add. It will also work even if some or all documents are missing the given keys so it's very robust.

Full dataset dump from mongo for reference:

```
/* 1 */
{
    "_id" : ObjectId("5668ba739f2e3d0001a9af59"),
    "comment" : "Priority 2",
    "actions" : [ 
        "*"
    ],
    "match" : {
        "account" : "admin"
    },
    "priority" : 2
}

/* 2 */
{
    "_id" : ObjectId("5668ba739f2e3d0001a9af5d"),
    "comment" : "Priority 1",
    "actions" : [ 
        "*"
    ],
    "match" : {
        "account" : "/.+/",
        "name" : "${account}/*"
    },
    "priority" : 1
}

/* 3 */
{
    "_id" : ObjectId("5671e9d1f62858550879b8bc"),
    "comment" : "Apriority 1",
    "actions" : [ 
        "*"
    ],
    "match" : {
        "account" : "/.+/",
        "name" : "${account}/*"
    },
    "priority" : 1
}

/* 4 */
{
    "_id" : ObjectId("5671ea04f62858550879b8bd"),
    "comment" : "priority none",
    "actions" : [ 
        "*"
    ],
    "match" : {
        "account" : "/.+/",
        "name" : "${account}/*"
    }
}

/* 5 */
{
    "_id" : ObjectId("5671eaa2f62858550879b8be"),
    "comment" : "APriority 0",
    "actions" : [ 
        "*"
    ],
    "match" : {
        "account" : "admin"
    },
    "priority" : 0
}
```

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/57)

### DIFF
--- a/auth_server/authz/acl_mongo.go
+++ b/auth_server/authz/acl_mongo.go
@@ -1,6 +1,7 @@
 package authz
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -10,6 +11,13 @@ import (
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 )
+
+type MongoACL []MongoACLEntry
+
+type MongoACLEntry struct {
+	ACLEntry `bson:",inline"`
+	Seq      *int
+}
 
 type ACLMongoConfig struct {
 	MongoConfig *mgo_session.Config `yaml:"dial_info,omitempty"`
@@ -28,9 +36,9 @@ type aclMongoAuthorizer struct {
 	CacheTTL         time.Duration `yaml:"cache_ttl,omitempty"`
 }
 
-// NewACLMongoAuthorizer creates a new ACL Mongo authorizer
+// NewACLMongoAuthorizer creates a new ACL MongoDB authorizer
 func NewACLMongoAuthorizer(c *ACLMongoConfig) (Authorizer, error) {
-	// Attempt to create new mongo session.
+	// Attempt to create new MongoDB session.
 	session, err := mgo_session.New(c.MongoConfig)
 	if err != nil {
 		return nil, err
@@ -67,7 +75,7 @@ func (ma *aclMongoAuthorizer) Authorize(ai *AuthRequestInfo) ([]string, error) {
 // Validate ensures that any custom config options
 // in a Config are set correctly.
 func (c *ACLMongoConfig) Validate(configKey string) error {
-	//First validate the mongo config.
+	//First validate the MongoDB config.
 	if err := c.MongoConfig.Validate(configKey); err != nil {
 		return err
 	}
@@ -119,7 +127,7 @@ func (ma *aclMongoAuthorizer) continuouslyUpdateACLCache() {
 
 func (ma *aclMongoAuthorizer) updateACLCache() error {
 	// Get ACL from MongoDB
-	var newACL ACL
+	var newACL MongoACL
 
 	// Copy our session
 	tmp_session := ma.session.Copy()
@@ -128,13 +136,43 @@ func (ma *aclMongoAuthorizer) updateACLCache() error {
 	defer tmp_session.Close()
 
 	collection := tmp_session.DB(ma.config.MongoConfig.DialInfo.Database).C(ma.config.Collection)
-	err := collection.Find(bson.M{}).All(&newACL)
-	if err != nil {
+
+	// Create sequence index obj
+	index := mgo.Index{
+		Key:      []string{"seq"},
+		Unique:   true,
+		DropDups: false, // Error on duplicate key document instead of drop.
+	}
+
+	// Enforce a sequence index. This is fine to do frequently per the docs:
+	// https://godoc.org/gopkg.in/mgo.v2#Collection.EnsureIndex:
+	//    Once EnsureIndex returns successfully, following requests for the same index
+	//    will not contact the server unless Collection.DropIndex is used to drop the same
+	//    index, or Session.ResetIndexCache is called.
+	if err := collection.EnsureIndex(index); err != nil {
 		return err
 	}
+
+	// Get all ACLs that have the required key
+	if err := collection.Find(bson.M{}).Sort("seq").All(&newACL); err != nil {
+		return err
+	}
+
 	glog.V(2).Infof("Number of new ACL entries from MongoDB: %d", len(newACL))
 
-	newStaticAuthorizer, err := NewACLAuthorizer(newACL)
+	// It is possible that the top document in the collection exists with a nil Seq.
+	// if that's true we pull it out of the slice and complain about it.
+	if len(newACL) > 0 && newACL[0].Seq == nil {
+		topACL := newACL[0]
+		return errors.New(fmt.Sprintf("Seq not set for ACL entry: %+v", topACL))
+	}
+
+	var retACL ACL
+	for _, e := range newACL {
+		retACL = append(retACL, e.ACLEntry)
+	}
+
+	newStaticAuthorizer, err := NewACLAuthorizer(retACL)
 	if err != nil {
 		return err
 	}
@@ -144,7 +182,7 @@ func (ma *aclMongoAuthorizer) updateACLCache() error {
 	ma.staticAuthorizer = newStaticAuthorizer
 	ma.lock.Unlock()
 
-	glog.V(2).Infof("Got new ACL from MongoDB: %s", newACL)
-	glog.V(1).Infof("Installed new ACL from MongoDB (%d entries)", len(newACL))
+	glog.V(2).Infof("Got new ACL from MongoDB: %s", retACL)
+	glog.V(1).Infof("Installed new ACL from MongoDB (%d entries)", len(retACL))
 	return nil
 }
--- a/docs/Backend_MongoDB.md
+++ b/docs/Backend_MongoDB.md
@@ -38,15 +38,20 @@ MongoDB quite easily. Below you can find a list of ACL entries that are ready to
 be imported into MongoDB. Those ACL entries reflect what's specified in the
 `example/reference.yml` file under the `acl` section (aka static ACL).
 
+The added field of seq is used to provide a reliable order which MongoDB does not
+guarantee by default, i.e. [Natural Sorting](https://docs.mongodb.org/manual/reference/method/cursor.sort/#return-natural-order).
+
+``seq`` is a required field in all MongoDB ACL documents. Any documents without this key will be excluded. seq uniqeness is also enforced.
+
 **reference_acl.json**
 
 ```json
-{"match" : {"account" : "admin"}, "actions" : ["*"], "comment" : "Admin has full access to everything."}
-{"match" : {"account" : "test", "name" : "test-*"}, "actions" : ["*"], "comment" : "User \"test\" has full access to test-* images but nothing else. (1)"}
-{"match" : {"account" : "test"}, "actions" : [], "comment" : "User \"test\" has full access to test-* images but nothing else. (2)"}
-{"match" : {"account" : "/.+/"}, "actions" : ["pull"], "comment" : "All logged in users can pull all images."}
-{"match" : {"account" : "/.+/", "name" : "${account}/*"}, "actions" : ["*"], "comment" : "All logged in users can push all images that are in a namespace beginning with their name"}
-{"match" : {"account" : "", "name" : "hello-world"}, "actions" : ["pull"], "comment" : "Anonymous users can pull \"hello-world\"."}
+{"seq": 10, "match" : {"account" : "admin"}, "actions" : ["*"], "comment" : "Admin has full access to everything."}
+{"seq": 20, "match" : {"account" : "test", "name" : "test-*"}, "actions" : ["*"], "comment" : "User \"test\" has full access to test-* images but nothing else. (1)"}
+{"seq": 30, "match" : {"account" : "test"}, "actions" : [], "comment" : "User \"test\" has full access to test-* images but nothing else. (2)"}
+{"seq": 40, "match" : {"account" : "/.+/"}, "actions" : ["pull"], "comment" : "All logged in users can pull all images."}
+{"seq": 50, "match" : {"account" : "/.+/", "name" : "${account}/*"}, "actions" : ["*"], "comment" : "All logged in users can push all images that are in a namespace beginning with their name"}
+{"seq": 60, "match" : {"account" : "", "name" : "hello-world"}, "actions" : ["pull"], "comment" : "Anonymous users can pull \"hello-world\"."}
 ```
 
 **Note** that each document entry must span exactly one line or otherwise the


rojer:
thanks for addressing the issue.

order is very important for ACLs, since we match until first match is found.
since mongo provides no guarantees about the order of documents in a collection, we have to introduce it. if it's not present, then the ACL may be completely wrong and security may be compromised (e.g. if we start matching a permissive entry first instead of a restrictive which is meant to be higher). so the parameter should not be optional, we should require it.

so, no objections on the approach, but two questions about the implementation:
1) do you envision any situation where it might be necessary to sort by more than one key?
2) is it important for the key to be configurable?

currently i don't see scenarios where more complex sorting would be needed, so i'd go with a mandatory field, presence of which should be checked when parsing. i'd also like it to be called "sequence" instead of priority.

carsonoid:
The only issue with making it mandatory is backwards compatibility. However if we make it so that we have ``["sequence "]`` as a default instead of simply breaking when the field is missing that would make sense. Having a multi-field sort is probably unnecessary 99% of the time but its very simple to implement so I just added it in there.  We could also optionally filter out any results without the sequence key but that would really break old installations.

So to summary I think my changes should be:

1. Change ``priority`` to ``sequence``
1. Default the sort_keys to be ``["sequence"]`` if the sort_keys configuration option is not given or is an empty list
1. Update the docs appropriately.

It's your project though so feel free to make the final call.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K5lvEfWrmLQQf1e9sxh:-1358662045)*

rojer:
in this case, the current behavior is obviously wrong. my bad for not catching it in review, my only lame excuse is that i was not familiar with mongo.
i think, in fact, that in this case breaking compatibility will attract attention to obviously wrong configurations and is thus justified. unfortunately, i don't have any communication channels to announce it upfront. i also don't know how many users there are, if any.
i bet original author will want to know, so @kwk - heads up.

with that, let's go ahead and add a required "sequence" field to the ACL stored in MongoDB. make sure its presence is verified on load and do not allow duplicates (as that, iiuc, will make the order unstable again).

rojer:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[auth_server/authz/acl_mongo.go, line 31 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DrYLC0zJn883eCrz:-K66DrYLC0zJn883eCs-:-1569417803) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/auth_server/authz/acl_mongo.go#L31)):*
s/var/const/

---

*[auth_server/authz/acl_mongo.go, line 139 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DCb9BNwE7HF1jBNS:-K66DCb9BNwE7HF1jBNT:1166480012) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/auth_server/authz/acl_mongo.go#L139)):*
simply excluding is bad, because it makes failure not obvious.
is there a way to ensure that all the documents in the collection have the field?

also, gofmt ALL the things.

---

*[docs/Backend_MongoDB.md, line 49 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DvCaaOQC9yiDMJrK:-K66DvCbsnFutnkmMqHU:508646991) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/docs/Backend_MongoDB.md#L49)):*
sequence is the field of the document, not the match section, right?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[auth_server/authz/acl_mongo.go, line 139 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DCb9BNwE7HF1jBNS:-K69tN_mMnYBqhgHE1ms:-1221009167) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/auth_server/authz/acl_mongo.go#L139)):*
MongoDB is a "schemaless" database so we can't require a field. The best we can do is ensure the index and control what we query.  Mandating a key has to be done by the process that inserts documents into the collection. Code that just uses the collection can't enforce anything.  I'll put a gofmt line in my pre-commit since I keep forgetting about it. Thanks!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
the part about enforcement by the DB i understand, but it doesn't mean we should simply ignore such entries.
can we fetch all the documents in the collection and manually verify that all of them have the sequence field?
we can have mongo perform the sorting by requesting sort function, but go over the result to make sure all the entries have distinct sequence numbers.

kwk:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[docs/Backend_MongoDB.md, line 42 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vNm-Y3Q7xiiP8O8O8:-K7vNm-Y3Q7xiiP8O8O9:-1803416106) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L42)):*
s/Mongo/MongoDB

---

*[docs/Backend_MongoDB.md, line 44 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vO2sgngv761lvncAA:-K7vO2sgngv761lvncAB:-558270722) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L44)):*
s/Mongo/MongoDB

Maybe we should mention that the sequence is not required to be 1, 2, 3, but instead can be 10, 100, 500, 4000 or whatever you like. Spacing in between the sequence might aid in manually integrating more ACL documents between two documents.

---

*[docs/Backend_MongoDB.md, line 49 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOckkXG34zYRKKniB:-K7vOckkXG34zYRKKniC:390681252) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L49)):*
Best start with 10

---

*[docs/Backend_MongoDB.md, line 50 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOfTzGtuX7mko3Tx9:-K7vOfTzGtuX7mko3TxA:-1303998820) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L50)):*
s/2/20

---

*[docs/Backend_MongoDB.md, line 51 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOh5l0W3fj7qFppPV:-K7vOh5l0W3fj7qFppPW:-1303968998) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L51)):*
s/3/30

---

*[docs/Backend_MongoDB.md, line 52 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOiYm3SLQjJdffz9c:-K7vOiYm3SLQjJdffz9d:-1303939176) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L52)):*
s/4/40

---

*[docs/Backend_MongoDB.md, line 53 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOjX-cpMwObRZ12w8:-K7vOjX-cpMwObRZ12w9:-1303909354) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L53)):*
s/5/50

---

*[docs/Backend_MongoDB.md, line 54 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOkad_88I4APAsBAQ:-K7vOkad_88I4APAsBAR:-1303879532) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L54)):*
s/6/60

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
+1 to what @kwk said, plus sequence validation from previous comment and we're good to go.

will you be able to finish this PR?

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K7ydhHTzSSga8mLV_pO:-315927228)*

carsonoid:
What should we do when a document doesn't have a sequence?

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K801M9-gpW5exDBUN6C:-547408386)*

rojer:
refuse to load whole ruleset. on startup it means - exit with failure, on background reload - loud complaints in the log.
i.e., act as we currently do on any other validation failure.

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K83kEwAUvtAcFuUKotT:1616346401)*

kwk:
@rojer That sounds good. In case a background reload fails and somebody want to authenticate/authorize I think the authentication/authorization should not be granted. What do you think? I haven't looked it up but that is what I would expect the server to do.

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K83mfBZWrHP1ZAtI-zs:-379166704)*

rojer:
no, what we do is we continue running on previous ("last known good") configuration and complain loudly in the logs. hopefully somebody is paying attention (fat chance, i know, but this is the best we can do :).

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K83wbUT9yvHlWn4hND3:-1461382541)*

carsonoid:
While coding this change I've discovered that manual duplicate sequence checking is basically pointless. Once the index is created in the collection it is not possible to insert a document with a duplicate sequence. The inserting process will error out. 

Furthermore, if you try to create the index on a collection that already has duplicate indexes the auth_server will error out on ensuring the index; which is what we want. An undefined(nil) sequence can only exist in one document without the ensureIndex call returning an error.

What this all means is that the following states are already handled by my existing code:

1.  The updated auth_server is run against an existing ACL collection where zero documents have a sequence field OR an existing ACL collection with 2 or more documents where some of the documents have a duplicate sequence field.
  * auth_server will refuse to run since it is unable to ensure the sequence index.
2. The updated auth_server is run against a new collection.
  * auth_server will ensure the index and it will not be possible to insert acl documents that have a duplicate sequence.  You **will** be able to insert one document without a sequence field but after that "null" will be considered used and no more documents without sequence will be insert able. This is all enforced by MongoDB and requires no code change beyond index enforcing. The null ACL entry will be filtered out of the results and can cause no harm.

I can keep my newly written code that checks for (1)nil sequence and (2)for duplicate sequences. However since item 1 can only happen once and item 2 can **never** happen it all seems moot at this point.

If we leaving the sequence checking and hinting as-is the worst thing that can possibly happen is that one document with a nil index is inserted and ignored.  We can crash and stop refreshing the ACL cache on that but that seems like it would be more damaging than to just ignore one ACL. Especially since acls are grant only so it's not like we are going to allow something that should have been blocked.

The one change that would make sense is to not filter the one non-sequenced doc out. But instead to log stating that it was ignored so that someone looking at the logs would know why their document was not being used. Again, this can only happen to one document in the collection.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K9T_F_OiBYrzVT2iZFw:-130856775)*

carsonoid:
*[auth_server/authz/acl_mongo.go, line 31 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DrYLC0zJn883eCrz:-K9Txc_ODndSvvFmB4SO:676635175) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/auth_server/authz/acl_mongo.go#L31)):*
Done

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[auth_server/authz/acl_mongo.go, line 139 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DCb9BNwE7HF1jBNS:-K9TySlmEg0jQYOXpsX2:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/auth_server/authz/acl_mongo.go#L139)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[docs/Backend_MongoDB.md, line 42 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vNm-Y3Q7xiiP8O8O8:-K9TyUHJ7z6VTMAW-bwY:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L42)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[docs/Backend_MongoDB.md, line 44 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vO2sgngv761lvncAA:-K9TyVCJDW0EtVXPnXcp:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L44)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[docs/Backend_MongoDB.md, line 49 \[r2\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K66DvCaaOQC9yiDMJrK:-K9TyZZ9W0rfJWDGR4Pw:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/ff6ae0b4447a7c97689535323e479b32941e40c3/docs/Backend_MongoDB.md#L49)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[docs/Backend_MongoDB.md, line 49 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOckkXG34zYRKKniB:-K9Ty_RpNszKMn_BhsLb:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L49)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions.

---

*[docs/Backend_MongoDB.md, line 50 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOfTzGtuX7mko3Tx9:-K9TyaABWqEFffWvK2AA:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L50)):*
Done.

---

*[docs/Backend_MongoDB.md, line 51 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOh5l0W3fj7qFppPV:-K9TyaMu2nlYJFmvDRPr:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L51)):*
Done.

---

*[docs/Backend_MongoDB.md, line 52 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOiYm3SLQjJdffz9c:-K9Tyabg_E4hUQFNjYNS:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L52)):*
Done.

---

*[docs/Backend_MongoDB.md, line 54 \[r3\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K7vOkad_88I4APAsBAQ:-K9TybOJW4M-aCNHLSVq:-499146009) ([raw file](https://github.com/cesanta/docker_auth/blob/b019f30eb0a58eda55b5113b560ef8b0b5451e9b/docs/Backend_MongoDB.md#L54)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
thanks for the detailed analysis!
wrt nil index - my preference is still to fail to load ACL if there is a doc with a nil index. that way it's easier to explain: documents must have sequence field, and that's that. no need to document that the nil-indexed acl will be ignored if present.

---

Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K9U1O2C9n1JJV16QXP7:-1216067759)*

carsonoid:
You got it. Newest push will log out the error and ignore new acls until the sequence is fixed.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-:-K9U6zInhi03BWfgpuNy:-280009164)*

rojer:
Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[auth_server/authz/acl.go, line 20 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd97d13HW89CnWUO6:-K9Yd97d13HW89CnWUO7:253120237) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl.go#L20)):*
sorry for last minute request, but let's call the JSON field `seq` for brevity.

also, since it's only used by Mongo, let's isolate it there: create struct MongoACLEntry that embeds ACLEntry and adds Sequence.

---

*[auth_server/authz/acl_mongo.go, line 159 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd1P1_F9wvEwU9HiS:-K9Yd1P1_F9wvEwU9HiT:1302642646) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl_mongo.go#L159)):*
just return an error here, it will be logged in the calling function

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[auth_server/authz/acl_mongo.go, line 159 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd1P1_F9wvEwU9HiS:-K9YugiURh6jk8b4dWH6:-1191888303) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl_mongo.go#L159)):*
Returning an error will cause the auth_server to die.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[auth_server/authz/acl.go, line 20 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd97d13HW89CnWUO6:-K9YzS2KmemcFS3Z-0eV:-566912294) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl.go#L20)):*
I'm not sure how to do that and still have the ``.All(&newACL)`` call work. 

When I do something like

```
type MongoACL []MongoACLEntry

type MongoACLEntry struct {
        Entry *ACLEntry
        Seq    *int             `yaml:"seq,omitempty"`
}

...

var newACL MongoACL

...
if err := collection.Find(bson.M{}).Sort("sequence").All(&newACL); err != nil {
	return err
}
```

When this returns, newACL is just a bunch of nils. How do I make it load into a nested struct?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

kwk:
Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[auth_server/authz/acl.go, line 20 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd97d13HW89CnWUO6:-K9alWyngVvj1S-39esI:-613248633) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl.go#L20)):*
Not quite sure but what if you *embed* the `ACLEntry` into the `MongoACLEntry`? 

(See this URL for embedding if not familiar with this concept: https://golang.org/doc/effective_go.html#embedding)

I think we initially had this for the Config struct [here](https://github.com/cesanta/docker_auth/blob/master/auth_server/mgo_session/mgo_session.go#L30)  where we embed the `DialInfo` struct. It is not present any longer.

```go
type MongoACL []MongoACLEntry

type MongoACLEntry struct {
// NOTE THE FOLLOWING LINE IS MISSING A NAME!!!
// You can access any ACLEntry element from a MongoACLEntry from the outside by using 
// var acl MongoACLEntry
// if acl.ACLEntry.AnyElementInside ...
        ACLEntry               `yaml:",inline"`                 // NOTE THE INLINE
        Seq    *int             `yaml:"seq,omitempty"`
}

...

var newACL MongoACL

...
// I think this code should work when embedding the `ACLEntry` into the `MongoACLEntry`
if err := collection.Find(bson.M{}).Sort("sequence").All(&newACL); err != nil {
    return err
}
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[auth_server/authz/acl.go, line 20 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd97d13HW89CnWUO6:-K9bTLJaneo0FUx1qcXx:1948777302) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl.go#L20)):*
yes, what Konrad said.

---

*[auth_server/authz/acl_mongo.go, line 159 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd1P1_F9wvEwU9HiS:-K9bTDy32vgJd7M20Zpi:372364756) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl_mongo.go#L159)):*
on startup - yes, and that's intentional.
on refresh - no, it will complain and use stale acls.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

carsonoid:
*[auth_server/authz/acl.go, line 20 \[r5\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9Yd97d13HW89CnWUO6:-K9iJpEiyKp4fUH8Bb-a:-871543415) ([raw file](https://github.com/cesanta/docker_auth/blob/03c0bac830769f573c560423ea043ad3fad97de1/auth_server/authz/acl.go#L20)):*
So this was do-able i just needed a bson asset tag.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
Review status: 0 of 2 files reviewed at latest revision, 11 unresolved discussions.

---

*[auth_server/authz/acl_mongo.go, line 167 \[r6\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9ifIMR-fcBpYfOtyzI:-K9ifIMR-fcBpYfOtyzJ:-1190045610) ([raw file](https://github.com/cesanta/docker_auth/blob/400d995bec3ebbe66f822ef12070ac1a7b0f7721/auth_server/authz/acl_mongo.go#L167)):*
no need to check here, just return error and `continuouslyUpdateACLCache` will print a message

---

*[docs/Backend_MongoDB.md, line 49 \[r6\]](https://reviewable.io:443/reviews/cesanta/docker_auth/57#-K9ifvcfXLAul11Cxw0r:-K9ifvcgKZx4xYl7IfEz:2080351567) ([raw file](https://github.com/cesanta/docker_auth/blob/400d995bec3ebbe66f822ef12070ac1a7b0f7721/docs/Backend_MongoDB.md#L49)):*
update example to use "seq"

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
Reviewed 1 of 2 files at r6, 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/57)*

rojer:
ok, this is good to go. assuming it works after all the changes, because i have no way to test.

thank you!

PR #68, opened 2016-02-26
Add support for multiple scopes

### LINKED ISSUE #67: Support requests for multiple scopes
Per docker/docker#20387:

> 1.  cesanta/docker_auth only understands a single scope query parameter, and the docker engine is requesting push/pull for the repository you are pushing to and pull access for the repository you're mounting the blob from. This means you request both scopes and the auth server only returns a token with the single repository scope.
> 
> ...
> 
> I can address issue number 2 so that it immediately falls back to pushing the blob on auth failure, specifically for the case of a cross-repo push, but I'd recommend filing an issue with cesanta/docker_auth to support this workflow.

rojer:
Fixes #67

[Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/68)

### DIFF
--- a/auth_server/server/server.go
+++ b/auth_server/server/server.go
@@ -33,17 +33,6 @@ import (
 	"github.com/golang/glog"
 )
 
-type AuthRequest struct {
-	RemoteAddr string
-	User       string
-	Password   authn.PasswordString
-	ai         authz.AuthRequestInfo
-}
-
-func (ar AuthRequest) String() string {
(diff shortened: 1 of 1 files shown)


mmikulicic:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/docker_auth/68)*

PR #70, opened 2016-03-09
Make ACLMongo omitempty for serialization

jgsqware:
Hello,

I updated the config struct to ommit ACLMongo if empty, to avoid having `ACLMongo:  null` on serialization.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/70)


rojer:
thank you!

PR #81, opened 2016-03-25
example acl for catalog

dpetzel:
This took me a good deal of effort to figure out (thanks for your suggestions in other threads about `-v 2`). I figured I'd kick back what worked for me in the event it helps same someone else in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/81)


rojer:
thank you!

PR #104, opened 2016-06-06
Handle "invalid credentials" as a non-error

rojer:
It's kind of expected, we should return 403 in this case and not 500.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/104)

### DIFF
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -75,10 +75,16 @@ func (la *LDAPAuth) Authenticate(account string, password PasswordString) (bool,
 	if uSearchErr != nil {
 		return false, uSearchErr
 	}
+	if accountEntryDN == "" {
+		return false, nil // User does not exist
+	}
 	// Bind as the user to verify their password
 	if len(accountEntryDN) > 0 {
 		err := l.Bind(accountEntryDN, string(password))
 		if err != nil {
+			if ldap.IsErrorWithCode(err, ldap.LDAPResultInvalidCredentials) {
+				return false, nil
+			}
 			return false, err
 		}
 	}
@@ -173,8 +179,10 @@ func (la *LDAPAuth) ldapSearch(l *ldap.Conn, baseDN *string, filter *string, att
 		return "", err
 	}
 
-	if len(sr.Entries) != 1 {
-		return "", fmt.Errorf("User does not exist or too many entries returned.")
+	if len(sr.Entries) == 0 {
+		return "", nil // User does not exist
+	} else if len(sr.Entries) > 1 {
+		return "", fmt.Errorf("Too many entries returned.")
 	}
 
 	var buffer bytes.Buffer


ryarnyah:
+1

rojer:
@ryarnyah does that mean that you've tested it and it works?

ryarnyah:
Yes it works but if you have a nonexistent login it's not:

```bash
$ docker login 0.0.0.0:5000
Username: professor
Password: 
Error response from daemon: Get http://0.0.0.0:5000/v2/: unauthorized: authentication required
$ docker login 0.0.0.0:5000
Username: test
Password: 
Error response from daemon: Get http://0.0.0.0:5000/v2/: received unexpected HTTP status: 500 Internal Server Error
```

Test env:
```bash
$ cat docker-compose.yml 
version: '2'

services:
  ldap:
    image: rroemhild/test-openldap
    ports:
      - "389:389"
    networks:
      app_net:
        ipv4_address: 172.16.238.10

  dockerauth:
    image: cesanta/docker_auth:ldap
    ports:
      - "5001:5001"
    volumes:
      - /home/ryarnyah/tmp/docker_auth/test:/config:ro
      - /home/ryarnyah/tmp/docker_auth/test:/logs
      - /home/ryarnyah/tmp/docker_auth/test:/ssl
    command: /config/auth_config.yml
    restart: always
    links:
      - ldap:ldap
    networks:
      app_net:
        ipv4_address: 172.16.238.11

  registry:
    image: registry:2
    ports:
      - "5000:5000"
    volumes:
      - /home/ryarnyah/tmp/docker_auth/test:/ssl
      - /home/ryarnyah/tmp/docker_auth/test:/var/lib/registry
    restart: always
    links:
      - dockerauth:auth.docker
    environment:
      - REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/var/lib/registry
      - REGISTRY_AUTH=token
      - REGISTRY_AUTH_TOKEN_REALM=https://0.0.0.0:5001/auth
      - REGISTRY_AUTH_TOKEN_SERVICE="Docker registry"
      - REGISTRY_AUTH_TOKEN_ISSUER="Auth Service"
      - REGISTRY_AUTH_TOKEN_ROOTCERTBUNDLE=/ssl/server.pem
    networks:
      app_net:
        ipv4_address: 172.16.238.12
networks:
  app_net:
    driver: bridge
    driver_opts:
      com.docker.network.enable_ipv6: "true"
    ipam:
      driver: default
      config:
      - subnet: 172.16.238.0/24
        gateway: 172.16.238.1
      - subnet: 2001:3984:3989::/64
        gateway: 2001:3984:3989::1

$ cat auth_config.yml
server:  # Server settings.
  # Address to listen on.
  addr: ":5001"
  # TLS certificate and key.
  certificate: "/ssl/server.pem"
  key: "/ssl/server.key"

token:  # Settings for the tokens.
  issuer: "Auth Service"  # Must match issuer in the Registry config.
  expiration: 900


# Static user map.
ldap_auth:
  # Addr is the hostname:port or ip:port
  addr: 172.16.238.10:389
  # Setup tls connection method to be
  # "" or "none": the communication won't be encrypted
  # "always": setup LDAP over SSL/TLS
  # "starttls": sets StartTLS as the encryption method
  tls: none
  # set to true to allow insecure tls
  insecure_tls_skip_verify: false
  # In case bind DN and password is required for querying user information,
  # specify them here. Plain text password is read from the file.
  bind_dn: cn=admin,dc=planetexpress,dc=com
  bind_password_file: /ssl/password
  # User query settings. ${account} is expanded from auth request
  base: ou=people,dc=planetexpress,dc=com
  filter: (&(uid=${account})(objectClass=inetOrgPerson))

acl:
  # Admin has full access to everything.
  - match: {account: "/.+/"}
    actions: ["*"]

$ cat password
GoodNewsEveryone

```

Maybe in ldap_auth.go we need to change this to return empty buffer with no error(not tested):
```go
if len(sr.Entries) != 1 {
	return "", fmt.Errorf("User does not exist or too many entries returned.")
}
```

rojer:
@ryarnyah yup, i think you're right. amended the PR, built and pushed test97. can you try again?

auth_server/authn/ldap_auth.go
```diff
@@ -173,7 +179,9 @@ func (la *LDAPAuth) ldapSearch(l *ldap.Conn, baseDN *string, filter *string, att
 		return "", err
 	}
 
-	if len(sr.Entries) != 1 {
+	if len(sr.Entries) == 0 {
+		return "", nil // User does not exist
+	} else if len(sr.Entries) > 1 {
 		return "", fmt.Errorf("User does not exist or too many entries returned.")
```
ryarnyah: In this case, i think we need to change the error message : "User does not exist or too many entries returned." To "Too many entries returned."

ryarnyah:
@rojer I will test it tomorrow.

ryarnyah:
It's seem good for me :
```bash
$ docker login 0.0.0.0:5000
Username: test
Password: 
Error response from daemon: Get http://0.0.0.0:5000/v2/: unauthorized: authentication required
$ docker login 0.0.0.0:5000
Username: professor
Password: 
Error response from daemon: Get http://0.0.0.0:5000/v2/: unauthorized: authentication required
```

auth_server/authn/ldap_auth.go
rojer: you're right, amended

rojer:
ok, merging. thanks @ryarnyah !

PR #105, opened 2016-06-22
Add support for LetsEncrypt.

cellofellow:
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/105)

### DIFF
--- a/auth_server/Dockerfile
+++ b/auth_server/Dockerfile
@@ -1,4 +1,4 @@
-FROM busybox
+FROM golang
 EXPOSE 5001
 ENTRYPOINT ["/auth_server"]
 CMD ["/config/auth_config.yml"]
--- a/examples/reference.yml
+++ b/examples/reference.yml
@@ -15,6 +15,10 @@ server:  # Server settings.
   # TLS certificate and key.
   certificate: "/path/to/server.pem"
   key: "/path/to/server.key"
+  # Alternately, you may use LetsEncrypt
+  letsencrypt:
+    cachefile: "/path/to/cachefile.letsencrypt"
+    email: "your@address.email"
   # Take client's address from the specified HTTP header instead of connection.
   # May be useful if the server is behind a proxy or load balancer.
   # If configured, this header must be present, requests without it will be rejected.
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/facebookgo/httpdown"
 	"github.com/golang/glog"
 	fsnotify "gopkg.in/fsnotify.v1"
+	"rsc.io/letsencrypt"
 )
 
 type RestartableServer struct {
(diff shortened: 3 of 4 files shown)


rojer:
oh, nice, thanks for this!

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[auth_server/main.go, line 85 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/105#-KKuADLEOyUGgCRubw6j:-KKuADLEOyUGgCRubw6k:156056707) ([raw file](https://github.com/cesanta/docker_auth/blob/d6658ca0c823f871d525c47f8f56596a46a5488f/auth_server/main.go#L85)):*
> ```Go
> 		}
> 		if !m.Registered() {
> 			m.Register(c.Server.LetsEncrypt.Email, nil)
> ```

handle return value from Register here

---

*[auth_server/main.go, line 87 \[r1\]](https://reviewable.io:443/reviews/cesanta/docker_auth/105#-KKuAI_Soc6Rqip5Zvuj:-KKuAI_Soc6Rqip5Zvuk:-740776332) ([raw file](https://github.com/cesanta/docker_auth/blob/d6658ca0c823f871d525c47f8f56596a46a5488f/auth_server/main.go#L87)):*
> ```Go
> 			m.Register(c.Server.LetsEncrypt.Email, nil)
> 		}
> 		tlsConfig = &tls.Config{
> ```

to reduce the amount of c&p, move setting TLS parameters into a helper function

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cesanta/docker_auth/105#-:-KKuAAghYDLSPme7jRkG:-1074837448)*

raphink:
Cool beans. Any news on this?

SydOps:
@cellofellow 

Could you give a sample on how to prepare below cachefile?

    cachefile: "/path/to/cachefile.letsencrypt"

cellofellow:
How would you like the example? It just needs to be a path that the process can write to, creating a file holding keys and login information.

auth_server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM busybox
+FROM golang
```
rojer: please leave this change out

examples/reference.yml
```diff
@@ -15,6 +15,10 @@ server:  # Server settings.
   # TLS certificate and key.
   certificate: "/path/to/server.pem"
   key: "/path/to/server.key"
+  # Alternately, you may use LetsEncrypt
```
rojer: to clarify - in this case, using separate token.{certificate,key} is a must, yes?

rojer: also, for those not familiar with letsencrypt, give a link to the documentation.
the package doc (https://godoc.org/rsc.io/letsencrypt) gives a nice overview

tlvenn:
@cellofellow any update on this ?

techknowlogick:
For anyone coming here to use LE with this project, what I've done as a workaround is to use a self signed cert for the auth_server and then to put Caddy in front of auth server which handles LE nicely.

Example here: https://github.com/techknowlogick/docker-registry/blob/master/caddy/Caddyfile

rojer:
so, i have some personal experience with LE now.
LE is awesome, i'm a big fan! i promise some movement on this no later than i need to renew our own cert ;)

rojer:
this inspired https://github.com/cesanta/docker_auth/commit/a76426387685c721962e84e29f60c85456b28e86

thank you!

PR #114, opened 2016-07-13
Return NoMatch when user does not exist in LDAP

raphink:
The LDAP authn provider currently returns no error when the user is not found in LDAP.

For this reason, it fails authentication without checking the next authenticator, which is a wrong behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/114)

### DIFF
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -76,7 +76,7 @@ func (la *LDAPAuth) Authenticate(account string, password PasswordString) (bool,
 		return false, uSearchErr
 	}
 	if accountEntryDN == "" {
-		return false, nil // User does not exist
+		return false, NoMatch // User does not exist
 	}
 	// Bind as the user to verify their password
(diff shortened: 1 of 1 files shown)


rojer:
you're right, thanks!

raphink:
Thanks for merging. I found this bug while coding a new authn provider for GitHub, which I will submit very soon (so get ready for a large PR :smile: )

rojer:
good! this is something i've been meaning to do myself, but haven't got around to due to lacjk of time.
i will say upfront though that i'd like to avoid using github password for docker login.
if your solution uses plaintext gh password, an arrangement similar to what is done for google_auth should be used - i.e. temp password and token storage on backend.

raphink:
My solution uses the same system as the Google login, with a button on the https interface.

raphink:
Here's the PR: https://github.com/cesanta/docker_auth/pull/115

Issue #117, opened 2016-07-21
Teams ACLs


raphink:
As mentioned in https://github.com/cesanta/docker_auth/pull/115#issuecomment-234074177, once #115  is merged, it would be great to be able to use teams in the ACLS, something along the lines of:

```go
- match: {account: "/.+/", name: "${team}/*"}
  actions: ["push", "pull"]
 ```

where team would be matched against something like `regexp.QuoteMeta(strings.Join(ai.Teams))`.

As you mentioned in #115, it would be good if this was not specific to GitHub. Here's a suggestion. Currently, `AuthServer#Authenticate()` has the following signature:

```go
func (as *AuthServer) Authenticate(ar *authRequest) (bool, error) {
```

Instead of returning a boolean, we could make it return a structure containing information about the authenticated user, so any authn provider could provider that information.

```
type AuthAccount struct {
  Name   string
  Email   string
  Teams []string
}

func (as *AuthServer) Authenticate(ar *authRequest) (AuthAccount, error) {
  ...
}

func (as *AuthServer) doAuth(rw http.ResponseWriter, req *http.Request) {
  ...
  		ares, err = as.Authorize(ar, authnResult)
}
```

and `authz.go` could thus access this information from the authn provider, and filter on Email, Teams, etc.

What do you think of this architecture?

rojer:
@raphink so, i found a bit of free time over the weekend and put together #139 
with that, all you need is add some code to github_auth to assign gh_team labels.
label is a string -> []string map, so multiple teams will be ok and matching will check all of them.

PR #118, opened 2016-07-21
Teams ACLs

raphink:
This PR aims to implement #117

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/118)

### DIFF
--- a/auth_server/authn/authn.go
+++ b/auth_server/authn/authn.go
@@ -25,7 +25,7 @@ type Authenticator interface {
 	// A special NoMatch error is returned if the authorizer could not reach a decision,
 	// e.g. none of the rules matched.
 	// Implementations must be goroutine-safe.
-	Authenticate(user string, password PasswordString) (bool, error)
+	Authenticate(user string, password PasswordString) (*AuthUser, error)
 
 	// Finalize resources in preparation for shutdown.
 	// When this call is made there are guaranteed to be no Authenticate requests in flight
@@ -34,6 +34,12 @@ type Authenticator interface {
 
 	// Human-readable name of the authenticator.
 	Name() string
+}
+
+type AuthUser struct {
+	Name  string
+	Email string
+	Teams []string
 }
 
 var NoMatch = errors.New("did not match any rule")
--- a/auth_server/authn/ext_auth.go
+++ b/auth_server/authn/ext_auth.go
@@ -77,7 +77,7 @@ func NewExtAuth(cfg *ExtAuthConfig) *extAuth {
 	return &extAuth{cfg: cfg}
 }
 
-func (ea *extAuth) Authenticate(user string, password PasswordString) (bool, error) {
+func (ea *extAuth) Authenticate(user string, password PasswordString) (*AuthUser, error) {
 	cmd := exec.Command(ea.cfg.Command, ea.cfg.Args...)
 	cmd.Stdin = strings.NewReader(fmt.Sprintf("%s %s", user, string(password)))
 	_, err := cmd.Output()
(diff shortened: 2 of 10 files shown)


rojer:
while what you did here is not terrible, i'd like a more generic mechanism of passing data between authn and authz parts. "team" is too specific to GH to deserve a field in the ACL struct. we don't want to be too specific.
perhaps, i'm thinking, we can make it possible for the authn mechanism to add _labels_, with values, and then teach authz/acl.go to match them. GH authn provider will pass down "team" label, GA may pass "domain", etc.

let's finish GH authn first and then get back to this.

raphink:
Dockerhub has [orgs and teams](https://docs.docker.com/docker-hub/orgs/), so it could be interesting to try and match that behavior maybe.

raphink:
As a note also, these informations (Domain/Organizations/Teams/Email) usually require to use the access token (both for Google and Github). So if we use this information for authz, we must either:

* always use the token, which kind of makes the expires_in/revalidate_after parameters moot
* or store the information (in a levelDB for example) until the token expires/needs revalidation, which means there is a delay in authorization propagation when a user's email/team/org is modified in Google/GitHub

rojer:
storing alongside ticket is fine. those who are concerned with propagation delay should be equally concerned with ticket staleness, they can reduce it.

raphink:
OK. Considering this problem is apparently linked to OAuth authentications, how about adding these fields to the `TokenDBValue`?

rojer:
so, i'm leaning towards generic mechanism for passing labels from authenticator (whichever authenticated the user) to authorizer. a string:string map, or maybe think ahead and use some sort of variant structure that can hold numbers for easy matching of numbers. i don't know what the situation is in Go-land, but in C++ and Qt we have QVariant - something of that sort would be ideal. but if it's too hairy, then we can start with just string values.

adding to tokendb - yeah, that'll do. how authn provider comes up with labels is really up to it, GH can store in tokendb. but again - make tokendb store generic labels. "team", "organization" and such should not leave GH authn code.
now, best syntax for matching labels in acls? ideally, as you suggested, it would be nice if labels just augmented the key space created by ticket so that we can do `match: {team: ...}` the same way as we do `match: {account: ...}`. but if it's too difficult to do, then something like `match: {account: ..., labels: {team: ...}}` is also an option.

raphink:
OK, so maybe as you suggest a `Labels` field as a `map[string]interface{}` in `TokenDBValue`. It would need to be a map of `interface{}` as there may be lists (e.g. teams) instead of strings (e.g. email) as values.

As for ACLs, as far as I'm concerned, I'm currently mostly interested in `match: { name: "${team}/*"}` (which is what I have implemented in this PR) to implement "team namespaces", but that comes as a bundle with the rest imo.

rojer:
not a fan of maps-to-interface{}, this should be the last resort.
my Go-fu is not that strong to begin with, and it's very late here, but please look around to somehow limit it to "either string or a list of strings" on entry.

rojer:
by "on entry" i mean so that type of label is checked at compile time rather than at run time with a type switch.

raphink:
Right. I'm not a Go expert either, but will try to get something strict-y.

raphink:
Would it work if we considered that every label has a slice of string as a value, and some happen to only contain one value in the slice?

raphink:
`${<label>}` would be replaced by the values joined with `|`.

carsonoid:
The recent addition of label support to mongo and static auth privides for a more generic way to do this.

techknowlogick:
Thanks for this PR :)

The codebase has evolved since this PR was made, and so I will close it. As mentioned above, I think label support would be a generic way to go about this.

PR #137, opened 2016-10-01
Add positional argument for real ip header.

jrodbeta:
Going to submit a few PRs for stuff we've found useful.

Currently the real IP header functionality defaults to grabbing the first IP address.  This would put responsibility of protecting the header upstream.  By adding a position argument I can easily select the trusted position of the header.

For example, we have this behind a google load balancer which provides the client IP for us in `X-Forwarded-For`: https://cloud.google.com/compute/docs/load-balancing/http/

The load balancer provides a header in the format of: `X-Forwarded-For: <client IP(s)>, <global forwarding rule external IP>`

Something like `X-Forwarded-For: 104.196.118.51, 130.211.9.23` is received where `104.196.118.51` is the client IP, `130.211.9.23` is the load balancer IP.

Now let's say that I attempt to spoof the header via curl:

```
curl -H "X-Forwarded-For: 10.0.0.1" https://registry
```

then the header turns into `X-Forwarded-For: 10.0.0.1, 104.196.118.51, 130.211.9.23`.  By default, reg-auth would just the provided IP in curl.

By defining a position that takes negative values (in this case -2), we were easily able to select the trusted IP regardless of what the client provided without needing more complexity upstream.

That's a lot, hope it makes sense.  More to come.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/137)

### DIFF
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -47,6 +47,7 @@ type Config struct {
 type ServerConfig struct {
 	ListenAddress string `yaml:"addr,omitempty"`
 	RealIPHeader  string `yaml:"real_ip_header,omitempty"`
+	RealIPPos     int    `yaml:"real_ip_pos,omitempty"`
 	CertFile      string `yaml:"certificate,omitempty"`
 	KeyFile       string `yaml:"key,omitempty"`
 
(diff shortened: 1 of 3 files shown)


rojer:
makes sense, thank you!

PR #138, opened 2016-10-01
Always prompt consent when signing in to Google OAuth

jrodbeta:
If you share the OAuth consent screen with other internal tools, you're unable use the Google login for this.

The current implementation assumes a refresh token always, but for already validated consents no refresh token is provided so Google login will fail.  More about this here:  http://stackoverflow.com/questions/10827920/not-receiving-google-oauth-refresh-token

This is easily fixed by always prompting for consent on the Google login (this PR).  However this could also be fixed by refactoring how the current Google auth functionality works (a larger effort).

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/138)


rojer:
thank you!

Issue #142, opened 2016-10-19
Support for external script for authorization


ceecko:
It would be great if we could call an external script to get rules for authorization. Similar to external authentication, the script would return json rules separated by a new line which would be then parsed statically and refreshed periodically similar to acl_mongo.

techknowlogick:
This can now be closed per merged PR #143

PR #143, opened 2016-10-21
Support for external authorization

ceecko:
PR for #142 
Added support to authorize request via an external script. The script returns a simple yes/no for the authz request through exit codes similarly to `ext_auth`

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/143)

### DIFF
--- a/auth_server/authz/acl_ext.go
+++ b/auth_server/authz/acl_ext.go
@@ -1,0 +1,99 @@
+/*
+   Copyright 2016 Cesanta Software Ltd.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       https://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package authz
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"strings"
+	"syscall"
+
+	"github.com/golang/glog"
+)
+
+type ACLExtConfig struct {
+	Command string   `yaml:"command"`
+	Args    []string `yaml:"args"`
+}
+
+type ACLExtStatus int
+
+const (
+	ACLExtAllowed ACLExtStatus = 0
+	ACLExtDenied  ACLExtStatus = 1
+	ACLExtError   ACLExtStatus = 2
+)
+
+func (c *ACLExtConfig) Validate() error {
+	if c.Command == "" {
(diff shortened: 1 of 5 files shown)


examples/reference.yml
```diff
@@ -225,3 +225,10 @@ acl_mongo:
   # the MongoDB server.
   # (See https://golang.org/pkg/time/#ParseDuration for a format description.)
   cache_ttl: "1m"
+
+# External authorization - call an external progam to authorize user.
+# JSON of authz.AuthRequestInfo is passed to command's stdin and exit code is examined.
+# 0 - allow, 1 - deny, other - error.
+acl_ext:
```
rojer: since it doesn't provide external _acls_, but external _authz_, let's call it `ext_authz`.

i have been thinking on sourcing acls externally as well. perhaps `acl_ext` should become an umbrella for acl sources, the thing you asked for in the first place, plus caching and refresh mechanism. so, `acl_ext` would configure one or more sources of acls (mongo, mysql, static file, ...), poll and cache them. but that's another story. let's keep this PR just about authz for now.

auth_server/authz/acl_ext.go
```diff
@@ -0,0 +1,99 @@
...
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"strings"
+	"syscall"
+
+	"github.com/golang/glog"
+)
+
+type ACLExtConfig struct {
```
rojer: ACl -> Authz, as noted below

ceecko:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[examples/reference.yml, line 232 at r1](https://reviewable.io:443/reviews/cesanta/docker_auth/143#-KUchZoa8P968dUjTlJH-r1-232:-KV9LnAcUk3tn7_E1jKz:b-dxcbts) ([raw file](https://github.com/cesanta/docker_auth/blob/912fd0da65e3956028daa5b12f2c93bca4810dce/examples/reference.yml#L232)):*
<details><summary><i>Previously, rojer wrote…</i></summary>
> since it doesn't provide external _acls_, but external _authz_, let's call it `ext_authz`.
> 
> 
> 
> i have been thinking on sourcing acls externally as well. perhaps `acl_ext` should become an umbrella for acl sources, the thing you asked for in the first place, plus caching and refresh mechanism. so, `acl_ext` would configure one or more sources of acls (mongo, mysql, static file, ...), poll and cache them. but that's another story. let's keep this PR just about authz for now.
</details>
Makes sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cesanta/docker_auth/143)*

rojer:
:lgtm:

---

Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cesanta/docker_auth/143#-:-KVHlSalAORyNFzwW6Gg:bnfp4nl)*

rojer:
thank you!

PR #162, opened 2017-01-17
Ext http authz

softprops:
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/162)


softprops:
sry about that. submitted pull against the wrong repo.

Issue #62, opened 2016-01-20
either ServerName or InsecureSkipVerify must be specified in the tls.Config


dpetzel:
I'm attempting to get LDAP auth working and I'm getting this error

```
authn #1 returned error: LDAP Result Code 200 "": tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
```

Things do work if I set `InsecureSkipVerify`, but for obvious reasons I'd rather not. I cam across [this](https://github.com/cesanta/docker_auth/issues/37#issuecomment-161865181) which suggests it may not be supported yet. 

I also seems I can not directly manipulate `tls.Config` Is it possible to set ServerName currently via configuration?

Thanks

PR #165, opened 2017-02-05
Add ServerName to tlsConfig when InsecureSkipVerify is false

dennisdegreef:
Should fix https://github.com/cesanta/docker_auth/issues/62

Related to https://github.com/cesanta/docker_auth/issues/37#issuecomment-161865181 and https://github.com/golang/go/issues/7342

For some reason, I get the following error, but that may be because I am trying to build master. Some futher insight is welcome. Should not be related to my change.
EDIT: After applying my patch to tag `1.2`, this builds cleanly.
```
F0205 02:16:59.049221    8077 main.go:171] Failed to load config: could not parse config: yaml: unmarshal errors:
  line 28: cannot unmarshal !!str `` into *authn.PasswordString
```

Any feedback is welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/165)

### DIFF
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -136,18 +136,25 @@ func (la *LDAPAuth) escapeAccountInput(account string) string {
 func (la *LDAPAuth) ldapConnection() (*ldap.Conn, error) {
 	var l *ldap.Conn
 	var err error
+
+	tlsConfig := &tls.Config{InsecureSkipVerify: true}
+	if !la.config.InsecureTLSSkipVerify {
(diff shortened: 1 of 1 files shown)


dennisdegreef:
It is nicer to have this fixed in go-ldap, which is something that is planned for v3.
See https://github.com/go-ldap/ldap/issues/94 and https://github.com/go-ldap/ldap/pull/65

kcd83:
I can confirm this works with our Active Directory, thanks!

For others who are also impatient and want this workaround
```
docker pull kcd83/docker_auth:addServerNameToTls
```

rojer:
thank you!

PR #139, opened 2016-10-09
Auth labels

rojer:
Add labels that can be passed from authentication to authorization rules.
Add ability to match them:
```
  acl:
   - match: {labels: {"foo": "bar", "baz": "/quux.*/"}}
```
All the label expressions must match ("and" logic on label expressions).
Labels are a string -> []string map, so a label can have multiple values
and all of them will be tried during matching, but only one is required
("or" logic on label values).

For now, only external authenticator can add labels, but in the future
e.g. GitHub authn can pass down GH organizations or teams the user
belongs to (should help #117).

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/139)

Issue #63, opened 2016-01-22
LDAP group matching example?


jmmills:
Can someone give me a hint on how I would write the config to source authentication from LDAP, and then specify access control based on group?

Essentially I want members of the developer group to be able to pull images but not push them (so nobody accidentally clobbers a production image).

Also, I was able to get registry v2 negotiating TLS and authentication with a basic auth backend, and it works like a charm. So thanks! (I didn't it using puppet and systemd unit files, if anyone is interested in example snippets of how that's done),

rojer:
i assume you want to control access to certain images based on group membership.
this is not possible currently and will require code changes - a "LDAP authorizer" will have to be written. currently we only have authenticator.

xinzhangcmu:
any plan to add support for group based authz? I have been trying the docker authz in our enterprise setting (and it works great, thanks!), and have the exact need

rojer:
no immediate plans. docker_auth is a side project for us at cesanta, and we don't use LDAP auth.
LDAP auth was contributed by @summerQLin - maybe you can ask (or pay) him to do it.
i will accept PR if you decide to do it and wish to contribute.

SydOps:
+1 Need this same feature as well. 

@summerQLin

Stephan1984:
+1 This feature would be great.
@summerQLin

tianon:
This seems kind of similar to https://github.com/cesanta/docker_auth/issues/117, and thus could probably be helped by https://github.com/cesanta/docker_auth/pull/139, right?

zapp42:
I think this issue can be closed.

PR #167, opened 2017-02-27
Map LDAP account attributes to labels such as groups #63

kcd83:
**Updated:**

I am proposing a configurable structure to leverage ldap account attributes as labels for access control. Let me know what you think.

I have added a `labels` mapping to the configuration, for example:
```
ldap_auth:
  ...
  filter: (&(|(userPrincipalName=${account})(userPrincipalName=${account}@example.com)(mail=${account}))(objectClass=person))
  labels:
    groups:
      attribute: memberOf
      parse_cn: true
    title:
      attribute: title
```

Can then be used with the current labels implementation like so:

```
 acl:
  - match: {labels: {"title": "Developer"}}
    actions: ["*"]
    comment: "If you call yourself a developer you can do anything."
  - match: {labels: {"groups": "Admin"}}
    actions: ["push"]
    comment: "If you are part of the admin group you can push."	
```
	
Using title for an acl is probably a really dumb idea. But it proves you can map arbitary attributes to arbitary labels. 

There is one special option, `parse_cn` which extracts the common name from a fully qualified value as found in `memberOf`.

**Original:**

I know little about LDAP and even less GO but I've taken the labels approach in #139 and populated groups to address #63.

```
  - match: {name: "docker_auth", labels: {"groups": "Admin"}}
    actions: ["*"]
    comment: "Admin users can do anything to docker_auth."
```

I'm simply looking through the account response for `memberOf` details. The alternative approach for ldap would be to look up groups and find member accounts.

A similar approach might be useful for #117

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/167)

### DIFF
--- a/auth_server/authn/ldap_auth.go
+++ b/auth_server/authn/ldap_auth.go
@@ -17,7 +17,6 @@
 package authn
 
 import (
-	"bytes"
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
@@ -71,10 +70,20 @@ func (la *LDAPAuth) Authenticate(account string, password PasswordString) (bool,
 	account = la.escapeAccountInput(account)
 
 	filter := la.getFilter(account)
-	accountEntryDN, uSearchErr := la.ldapSearch(l, &la.config.Base, &filter, &[]string{})
+
+	// dnAndGroupAttr := []string{"DN"} // example of no groups mapping attribute
+	groupAttribute := "memberOf"
+	dnAndGroupAttr := []string{"DN", groupAttribute}
+
+	entryAttrMap, uSearchErr := la.ldapSearch(l, &la.config.Base, &filter, &dnAndGroupAttr)
 	if uSearchErr != nil {
 		return false, nil, uSearchErr
 	}
+	if len(entryAttrMap) < 1 || entryAttrMap["DN"] == nil || len(entryAttrMap["DN"]) != 1 {
+		return false, nil, NoMatch // User does not exist
+	}
+
+	accountEntryDN := entryAttrMap["DN"][0]
 	if accountEntryDN == "" {
 		return false, nil, NoMatch // User does not exist
 	}
@@ -91,6 +100,20 @@ func (la *LDAPAuth) Authenticate(account string, password PasswordString) (bool,
 	// Rebind as the read only user for any futher queries
 	if bindErr := la.bindReadOnlyUser(l); bindErr != nil {
 		return false, nil, bindErr
+	}
+
+	// Extract group names from the attribute values
+	if entryAttrMap[groupAttribute] != nil {
+		rawGroups := entryAttrMap[groupAttribute]
+		labels := make(map[string][]string)
+		var groups []string
+		for _, value := range rawGroups {
+			cn := la.getCNFromDN(value)
+			groups = append(groups, cn)
+		}
+		labels["groups"] = groups
+
+		return true, labels, nil
 	}
 
 	return true, nil, nil
@@ -170,9 +193,9 @@ func (la *LDAPAuth) getFilter(account string) string {
 
 //ldap search and return required attributes' value from searched entries
 //default return entry's DN value if you leave attrs array empty
-func (la *LDAPAuth) ldapSearch(l *ldap.Conn, baseDN *string, filter *string, attrs *[]string) (string, error) {
+func (la *LDAPAuth) ldapSearch(l *ldap.Conn, baseDN *string, filter *string, attrs *[]string) (map[string][]string, error) {
 	if l == nil {
-		return "", fmt.Errorf("No ldap connection!")
+		return nil, fmt.Errorf("No ldap connection!")
 	}
 	glog.V(2).Infof("Searching...basedDN:%s, filter:%s", *baseDN, *filter)
 	searchRequest := ldap.NewSearchRequest(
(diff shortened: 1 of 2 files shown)


rojer:
@kcd83 thanks for the contribution! this looks very reasonable to me, but then again - i know very little about LDAP too.
have you actually used it yourself? if not, i'd like someone who actually uses LDAP to confirm that this works.

maybe @summerQLin can comment

kcd83:
Apparently modern AD installations/ configurations don't necessarily expose `memberof` and/or may prefer impersonation... anyone with ldap experience able to comment?

mrueg:
Is there any chance to merge this soon and split the AD part into a separate issue?

alonbl:
Hi,

I checked this on top of master and I can confirm it is working!

It is usable as-is so please merge it.

The problem with this approach is that the memberOf resolution does not support recursive groups. However, even with immediate groups there is a value.

Validated with AD.

I do need recursive groups so I ended up with the following filter instead of using this patch:
````
    (&(sAMAccountName=${account})(objectClass=person)(memberOf:1.2.840.113556.1.4.1941:=CN=XXX,OU=YYY,DC=X))
````
This is not flexible as I wish it to be... as I can use only single recursive group.

It is hard for me to imagine how in this generic attribute -> label approach a recursive solution may be built without code change. For active directory a list of filters to check recursive each group provides a resolution, for example the following will set in_group1 to "group1" if user is recursively in group1.

```
  filters:
    base:
      filter: <this is the base filter as currently implemented
    group1:
      filter: (&(sAMAccountName=${account})(objectClass=person)(memberOf:1.2.840.113556.1.4.1941:=CN=group1,OU=YYY,DC=X))
      labels:
        in_group1:
          attribute: cn
```
In other directories it is hard to perform this without recursion. As groups are so important, maybe a special behaviour is needed.

Thanks!

File: auth_server/authn/ldap_auth.go [hunk 1]
```diff
@@ -27,16 +26,20 @@ import (
...
-	GroupBaseDN           string `yaml:"group_base_dn,omitempty"`
-	GroupFilter           string `yaml:"group_filter,omitempty"`
+	Addr                  string              `yaml:"addr,omitempty"`
+	TLS                   string              `yaml:"tls,omitempty"`
+	InsecureTLSSkipVerify bool                `yaml:"insecure_tls_skip_verify,omitempty"`
+	Base                  string              `yaml:"base,omitempty"`
+	Filter                string              `yaml:"filter,omitempty"`
+	BindDN                string              `yaml:"bind_dn,omitempty"`
+	BindPasswordFile      string              `yaml:"bind_password_file,omitempty"`
+	LabelMaps             map[string]LabelMap `yaml:"labels,omitempty"`
```
alonbl: please add doc into examples/reference.yml otherwise nobody will know :)

mrueg:
@kcd83 any chance to update it to @alonbl's comments so it could be merged? :)

kcd83:
@mrueg @alonbl I have added the example configuration to reference.yml

I agree that additional queries of groups would be great but it isn't something I'll be able to look at. Mostly because this is all the Go I've ever written and I could barely remember how to build it :)

examples/reference.yml
```diff
@@ -134,6 +134,16 @@ ldap_auth:
...
+  # Labels can be mapped from LDAP attributes
+  labels:
+    # Add the user's title to a label called title
+    title:
+      attribute: title
+    # Add the user's memberOf values to a label called groups
+    groups:
+      attribute: memberOf
+      # Special handling to simplify the values to just the common name
+      parse_cn: true
```
alonbl: Good, I would also suggest to provide an example at acl section of how to use these attributes

alonbl:
Thanks @kcd83! As I wrote, this patch is helpful as-is, the group thing can be added later.

rojer:
please address @alonbl 's comments about examples and docs and this can be merged. thanks!

kcd83:
Whoops i was supposed to call that group not groups to match the existing acls. Would it be worth saying "see the group acls below" or similar also?

```
 - match: {labels: {"group": "VIP"}}
    actions: ["push"]
    comment: "Users assigned to group 'VIP' is able to push"
  - match: {labels: {"group": "/trainee|dev/"}}
    actions: ["push", "pull"]
    comment: "Users assigned to group 'trainee' and 'dev' is able to push and pull"
 

```

alonbl:
You gave a great example in your comment#0 :)
I tested based on that...

```
acl:
  - match: {labels: {"title": "Developer"}}
    actions: ["*"]
    comment: "If you call yourself a developer you can do anything."
  - match: {labels: {"groups": "Admin"}}
    actions: ["push"]
    comment: "If you are part of the admin group you can push."	
```

yes, a nice comment will be helpful as well.

auth_server/authn/ldap_auth.go
pinepain: It seems to be this is already added to `examples/reference.yml`

examples/reference.yml
pinepain: For those who are not very familiar with LDAP, it would be nice to provide example when `parse_cn` should be `true` and when `false` (false is default behavior, right)?

pinepain:
@kcd83 Thank you for this nice feature! 

Are there any estimates on when this could be good to merge and released? 

My use case is to authenticate and authorize users with [JumpCloud LDAP](https://support.jumpcloud.com/customer/en/portal/articles/2439911), the [Jenkins Integration Guide](https://support.jumpcloud.com/customer/en/portal/articles/2440013-configuring-jenkins-to-use-jumpcloud-s-ldap-as-a-service) has some examples to filter by user groups

examples/reference.yml
kcd83: It is a special case _only_ for `cn` fields, it isn't very reusable

kcd83:
I _think_ I addressed all the suggestions.

I don't think this is the place for an LDAP tutorial however, particularly because it is a very limited implementation. The examples can be used and someone with more ldap knowledge could improve this in time. 

Ok to merge?

examples/reference.yml
pinepain: That's just to get `foo` from `cn=foo,ou=bar,dc=bar`, right?

pcallewaert:
Hi,

This probably has no support for RFC 2307 style where group membership has members listed on the group object rather than using groups listed on user object?

pinepain:
@kcd83 Thanks for you hard work! Can't wait to have this changes in a release. Would you have some time to rebase your branch onto current master so that there will be no conflicts and to make it easier for @rojer to merge it by just pressing the merge button?

@rojer do you think this PR is good to be merged?

mrueg:
@kcd83 can you update this PR to remove the file conflicts?
@rojer I've been using this patch for quite a while, it works well. Can we get it merged?

aparamon:
This feature is desirable -- any obstacle to merge?

rojer:
closing in favor of https://github.com/cesanta/docker_auth/pull/228

PR #170, opened 2017-03-30
Fix checking of service field in ACL.

roman-vynar:
There is `service` field to define in the ACL, however the code misses to check it.
This PR fixes this so you can have the rules like this:
```
  - match: {account: "", service: "notary.quiq.sh"}
    actions: ["pull"]
    comment: "Anonymous users can get signatures from notary."
```

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/170)


rojer:
thanks Roman! looks good

Issue #181, opened 2017-06-08
Support URL prefix


alonbl:
Hi,

Current URLs are hardcoded, as in:
```
        case req.URL.Path == "/":
                as.doIndex(rw, req)
        case req.URL.Path == "/auth":
                as.doAuth(rw, req)
        case req.URL.Path == "/google_auth" && as.ga != nil:
                as.ga.DoGoogleAuth(rw, req)
        case req.URL.Path == "/github_auth" && as.gha != nil:
                as.gha.DoGitHubAuth(rw, req)
```

Can you please consider supporting a configuration option of URL prefix (default '/'), so that the service safely play behind a proxy?

Docker-distribution support this as well[1]
```
http:
  addr: localhost:5000
  prefix: /my/nested/registry/
```

Thanks!

[1] https://github.com/docker/distribution/blob/master/docs/configuration.md

rojer:
no objections. feel free to implement and send a PR, or wait until myself of someone else gets to it.

alonbl:
Great! I will try to submit my first go patch :)

alonbl:
Hi,
Finally I had time!
It[1] was trivial (I think), and working in our setup.
Thanks!

[1] https://github.com/cesanta/docker_auth/pull/185

carsonoid:
I think this issue can be closed now @alonbl

alonbl:
Hi,
Sure, it is working for me per master.
Thanks!

PR #185, opened 2017-06-30
Feature: URL Prefix support (#181)

alonbl:
Usable when service is behind a proxy.
Tested with docker-distribution and apache httpd.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/185)

### DIFF
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -48,6 +48,7 @@ type Config struct {
 
 type ServerConfig struct {
 	ListenAddress string            `yaml:"addr,omitempty"`
+	PathPrefix    string            `yaml:"path_prefix,omitempty"`
 	RealIPHeader  string            `yaml:"real_ip_header,omitempty"`
 	RealIPPos     int               `yaml:"real_ip_pos,omitempty"`
 	CertFile      string            `yaml:"certificate,omitempty"`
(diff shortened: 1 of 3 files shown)


alonbl:
Hi,
Is it ok?
Thanks!

auth_server/server/config.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 
 type ServerConfig struct {
 	ListenAddress string            `yaml:"addr,omitempty"`
+	UrlPrefix     string            `yaml:"prefix,omitempty"`
```
rojer: let's call it path prefix (because that's what it is).
```
PathPrefix     string            `yaml:"path_prefix,omitempty"`
```

rojer:
sorry, severe case of `ENOTIME`-itis. looks ok overall, let's just call it path prefix, since it's a prefix to the url's path component

auth_server/server/config.go
alonbl: Done

rojer: `PathPrefix` and `path_prefix` respectively, please

alonbl: Done?

rojer:
thank you!

PR #193, opened 2017-08-15
docker_auth/github: store tokens in google cloud storage

mgenov:
Added a new implementation of TokenDB that uses Google Cloud Storage as backend for storing of the tokens.

This makes container independent from the file system of the container or from the target system and is good alternative to mounted clustered file systems to container.

Docs examples are updated regarding this change as specifying token_db configuration now becomes alternate between local file and gcs.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/193)

### DIFF
--- a/auth_server/authn/github_auth.go
+++ b/auth_server/authn/github_auth.go
@@ -32,15 +32,21 @@ import (
 )
 
 type GitHubAuthConfig struct {
-	Organization     string        `yaml:"organization,omitempty"`
-	ClientId         string        `yaml:"client_id,omitempty"`
-	ClientSecret     string        `yaml:"client_secret,omitempty"`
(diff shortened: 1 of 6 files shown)


rojer:
this looks good, but can you please put GCS token DB impl into its own file? otherwise LG

mgenov:
np, later today I'll move it in separate file.

rojer:
also, since it's a sizeable contribution, please sign Cesanta CLA: https://docs.cesanta.com/contributors_la.shtml

rojer:
almost there. please rename to `tokendb_gcs.go`

mgenov:
done.

rojer:
thank you!

PR #194, opened 2017-08-30
Return labels from mongo users

carsonoid:
mongo auth will now include labels when they are part of an authenticated user's record.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/194)


carsonoid:
@rojer This should be a quick review/approve.

rojer:
thanks!

PR #199, opened 2017-09-25
Stop quoting meta chars in labels

carsonoid:
This small change allows for access changes to potentially be 100% user based when using labels to match names.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/199)


rojer:
makes sense, thanks!

PR #206, opened 2017-11-15
Fix Static labels and document supported backends

### LINKED ISSUE #202: are labels in YAML config file ACL specifications in the latest docker image
I just started using docker_auth today - worked quickly out of the box, that is until I tried to use the labels as exampled in https://github.com/cesanta/docker_auth/blob/master/docs/Labels.md, but I could not get them to work.  I tried both the stable (:1) and :1.3 tags of the docker image, but neither seemed to work.  I didn't dig any further, such as trying a source build, but will if someone can tell me that it should be working.  I can of course provide more details if needed as well.

carsonoid:
Fixes #202

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/206)

### DIFF
--- a/auth_server/authn/static_auth.go
+++ b/auth_server/authn/static_auth.go
@@ -23,6 +23,7 @@ import (
 
 type Requirements struct {
 	Password *PasswordString `yaml:"password,omitempty" json:"password,omitempty"`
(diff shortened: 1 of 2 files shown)


rojer:
thanks!

PR #209, opened 2017-12-05
Add a non-tls example config

tsl0922:
Adds an example config for #25.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/209)


rojer:
committing with minor edit. thanks!

Issue #210, opened 2017-12-25
seems like docker_auth and registry both can't work on same ip with letsencrypt


chrishecker:
After a bunch of debugging, it appears letsencrypt requires the requester to be able to reply on 443, so if you've got registry and docker_auth both using it for SSL certs you're out of luck?  I wonder if it would be possible to have docker_auth use the json file from registry's letsencrypt request, which would be one way to solve this problem?

Otherwise I think I'm going to have to either get a "real" ssl cert for this server, or set up letsencrypt outside of registry and docker_auth and have it fetch the certs on a cron job?  Hmm, or I guess I could let registry fetch the certs and then write a little cron job to suck them out of the letsencrypt.json file and restart the docker_auth container...

Any thoughts?

Thanks for making docker_auth, looks like it's going to be super useful once I get it to work.  :)

Chris

techknowlogick:
@chrishecker I've solved this by using caddy to deal with letsencrypt and have it proxy to the registry and auth server. If you want you can see how I have it set up here: https://github.com/techknowlogick/docker-registry

chrishecker:
So I didn't really like the idea of another heavyweight moving part like a proxy server in the mix, and so I went a little crazy and realized that I could use the busybox environment that docker_auth is in to run a shell script that can take the main registry letsencrypt.json file and blow it out to a key and pem file for docker_auth, and if you modify the auth_config.yml file there's already code in docker_auth to restart with the new key and pem. It needs a little cleaning up, but it seems like a pretty clean solution until docker_auth can just take the letsencrypt.json file in its config and parse it when it changes, which is the right solution. Seems like it would only be a few lines of code but I don't have a go dev env set up so I haven't messed with it yet.

I can clean this code up and post it if others are interested.  I'm not sure if the right way is to inject this thing into docker_auth or have it just be another container in a compose file and they share the json file and whatnot. I assume the latter but I'm new to this container thing.

Chris

chrishecker:
Okay, I just ahead and wrote the code for this in docker_auth, it's way cleaner that way.  Here's the pull request, and the comment mentions the issues (nothing serious, a bit of a bootstrap ordering, and a note about testing): https://github.com/cesanta/docker_auth/pull/211

PR #211, opened 2017-12-26
added config settings to share letsencrypt.json with registry so both can run on same machine

chrishecker:
As discussed here: https://github.com/cesanta/docker_auth/issues/210

The problem is you can't run docker_auth and registry on the same machine with letsencrypt without jumping through hoops because that service needs to talk to both services on port 443.  You can run a proxy like @techknowlogick mentions in that thread, or I wrote a simple bash script that can run on a timer and extract the key and certs from letsencrypt.json, but both of these are hacks (the proxy counts on the request URLs not overlapping, and the extractor runs on a sleep timer or else needs the busybox version of inotify). Clearly it'd be better to have this built into docker_auth, so I looked at the code and it was pretty simple to add, and I'd never played with Go and was curious, so I just did it.

The change is pretty simple, it adds a registryjson setting to the letsencrypt section, and if that's there, it adds the letsencrypt.json to the watched files (for when registry updates it) and extracts the key and cert directly.

The startup can be a little awkward if there's no letsencrypt.json, so I'd suggest getting your registry up and running and hit the /v2/ endpoint to get it to create the letsencrypt.json, then you can safely start up docker_auth.  I have a compose file that has them both in it with the right volume mappings and stuff, so I can add that if people are interested in an example.  You can get by with just touching letsencrypt.json to start up, but that makes docker_auth fail a bunch and it's not pretty in the logs, so better to just bootstrap ones. Also, be careful, I hit the letsencrypt rate limit while testing the startup cases!

Also, if you're going to test it by touching the letsencrypt.json file outside the contains while they're running, you need to be careful. For some reason the Go notify doesn't fire when you just touch the file, so you need to edit it and save, but you need to use an editor that overwrites the existing file (like emacs and I think nano), not deletes it and creates a new file (like vi) because then the inode is different and the mapping into the containers doesn't work (see this thread: https://github.com/moby/moby/issues/15793). But, the registry does the right thing and its changes get picked up, so that's just a testing weirdness.

Let me know if you've got questions, and I hope this can get integrated!
Chris

PS. The Makefile change snuck in there...the make target line is probably a good one to keep, but feel free to nuke the ca-bundle.crt change.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/211)

### DIFF
--- a/auth_server/Makefile
+++ b/auth_server/Makefile
@@ -1,5 +1,5 @@
 MAKEFLAGS += --warn-undefined-variables
-IMAGE ?= cesanta/docker_auth
+IMAGE ?= checker/docker_auth
 COMPRESS_BINARY ?= false
 CA_BUNDLE = /etc/ssl/certs/ca-bundle.crt
 VERSION = $(shell cat version.txt)
--- a/auth_server/Makefile
+++ b/auth_server/Makefile
@@ -1,7 +1,7 @@
 MAKEFLAGS += --warn-undefined-variables
 IMAGE ?= cesanta/docker_auth
 COMPRESS_BINARY ?= false
-CA_BUNDLE = /etc/ssl/certs/ca-certificates.crt
+CA_BUNDLE = /etc/ssl/certs/ca-bundle.crt
 VERSION = $(shell cat version.txt)
 
 BUILDER_IMAGE ?= golang:1.8.0-alpine
@@ -27,7 +27,7 @@ build:
 	CGO_ENABLED=0 go build -v -i --ldflags=--s
 
 ca-certificates.crt:
-	cp $(CA_BUNDLE) .
+	cp $(CA_BUNDLE) $@
 
 build-release: ca-certificates.crt
 	docker run --rm -v $(PWD)/..:/go/src/github.com/cesanta/docker_auth \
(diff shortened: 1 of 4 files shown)


rojer:
sorry for the delayed response - busy as usual, blah-blah.
now, i am very interested in this, since i also have a setup that has both registry and auth server on the same IP and don't want to throw proxy into the mix.
i'd like to make startup with missing json file less noisy though, without resorting to compose.
can you test for file existence and if it's non-existent or empty, wait, like, 10 seconds, checking every second, before bailing out.

chrishecker:
Just as a heads up, Let's Encrypt support with the main registry:2 is totally busted right now, sadly, and this patch will have to be updated when they (eventually?) fix it.  See these links:

   - https://github.com/docker/distribution/issues/1903
   - https://github.com/docker/distribution/issues/2545
   - https://community.letsencrypt.org/t/2018-01-11-update-regarding-acme-tls-sni-and-shared-hosting-infrastructure/50188
   - https://community.letsencrypt.org/t/tls-sni-challenges-disabled-for-most-new-issuance/50316

techknowlogick:
Thanks for this PR :)

I'm going through existing PRs, and cleaning things up. 

I think would be better solved by something like certmagic which allows configuration around which challenges are run, and could even be extended to using DNS challenges

PR #218, opened 2018-02-15
Small improvements to the build process

karmi:
These couple of patches implement some small, but significant improvements for building the Go binary from source, most notably the addition of the `make build-release-fast` task, which cuts the time needed more than by half, by skipping the pulling of dependencies. On a MacBook Air 2013, it takes about 30 seconds, as opposed to almost 2 minutes for the `make build-release` task.

The `make custom-image` task needs to be run first, so a customized image with all the dependencies is being built.

It also allows to override the location of the certificate bundle, in order to use a different one, eg. the [Mozilla bundle](https://curl.haxx.se/ca/cacert.pem).

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/218)

### DIFF
--- a/auth_server/Dockerfile
+++ b/auth_server/Dockerfile
@@ -1,4 +1,4 @@
-FROM busybox
+FROM alpine
 ADD auth_server /docker_auth/
 COPY ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 ENTRYPOINT ["/docker_auth/auth_server"]
--- a/auth_server/Dockerfile-builderimage
+++ b/auth_server/Dockerfile-builderimage
@@ -1,0 +1,4 @@
+#FROM busybox
+FROM golang:1.8.0-alpine
+RUN apk update && apk add git make py2-pip && \
+    pip install GitPython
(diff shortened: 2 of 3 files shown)


karmi:
@alex-mohemian, right!, multi-stage Docker build probably makes more sense. (I think this solution was chosen in order to be as little invasive as possible to the existing workflow.)

alex-mohemian:
@karmi I figured as much. Since it's a Go application, using the `scratch` image made the most sense to me. `Makefile`s seem so antiquated :) Especially for running Docker commands!

rojer:
@carsonoid is right. i'm all fo rmaking the build simpler, will take a look when i have time

tboerger:
> @karmi I figured as much. Since it's a Go application, using the `scratch` image made the most sense to me. `Makefile`s seem so antiquated :) Especially for running Docker commands!

No way! I need to build my own Docker image now because the image already provides too few tools installed if I want to use the external script feature...

If you try to execute wget (curl is not installed) to an https page it results in an error like:

```
Connecting to example.com (192.168.1.1:443)
wget: can't execute 'ssl_helper': No such file or directory
wget: error getting response: Connection reset by peer
```

wgerlach:
Using multi-stage builds make totally sense. As for the question what base image to use, the `scratch` image or a minimalistic image such as `busybox` or `alpine`, I would argue it makes sense to provide both on docker hub, for example `cesanta/docker_auth:lastest` and `cesanta/docker_auth:latest-alpine` (e.g. to support the external script feature).

techknowlogick:
Closing as I have just merged a multi-stage docker image

PR #220, opened 2018-02-15
Add a Redis-based token storage implementation

karmi:
This pull request adds a new type, `redisTokenDB`, which implements a token storage based on [_Redis_](https://redis.io).

The primary motivation is to allow the authentication server to run in a distributed environments, with multiple instances of the server proxied by a load balancer. The default file-based implementation is insufficient in such a scenario, because each server would have its own database of tokens, and the Google Cloud Storage-based implementation is prohibitive when the service would run eg. on AWS, due to latency issues.

The Redis-based implementation can work with a single Redis instance or a Redis Cluster — the corresponding client is selected based on the YAML configuration, where the user provides either a single `url` value as a string, or one or more `urls` as an array of strings (see the updated `reference.yml` file).

The Redis-based implementation adds two additional functions, `Encrypt()` and `Decrypt()`, which allow to encrypt the values before storing them, using the AES support in Go standard library, when the `encrypt_key` is set in the YAML configuration (again, see the updated `reference.yml` file).

The pull request also adds support for using the Redis-bases token storage in the Github authentication mechanism (`authn/github_auth.go`).

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/220)

### DIFF
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -196,8 +196,8 @@ func validate(c *Config) error {
 			return errors.New("github_auth.{client_id,client_secret,gcs_token_db{bucket,client_secret_file}} are required")
 		}
 
-		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.RedisTokenDB != nil && ghac.RedisTokenDB.NodeOptions.Addr == "" && len(ghac.RedisTokenDB.ClusterOptions.Addrs) < 1) {
-			return errors.New("github_auth.{client_id,client_secret,redis_token_db.{node_options,cluster_options}} are required")
+		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.RedisTokenDB != nil && ghac.RedisTokenDB.ClientOptions == nil && ghac.RedisTokenDB.ClusterOptions == nil) {
+			return errors.New("github_auth.{client_id,client_secret,redis_token_db.{redis_options,redis_cluster_options}} are required")
 		}
 
 		if ghac.HTTPTimeout <= 0 {
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -196,8 +196,8 @@ func validate(c *Config) error {
 			return errors.New("github_auth.{client_id,client_secret,gcs_token_db{bucket,client_secret_file}} are required")
 		}
 
-		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.RedisTokenDB != nil && ghac.RedisTokenDB.Url == "" && len(ghac.RedisTokenDB.Urls) < 1) {
-			return errors.New("github_auth.{client_id,client_secret,redis_token_db{url(s)}} are required")
+		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.RedisTokenDB != nil && ghac.RedisTokenDB.NodeOptions.Addr == "" && len(ghac.RedisTokenDB.ClusterOptions.Addrs) < 1) {
+			return errors.New("github_auth.{client_id,client_secret,redis_token_db.{node_options,cluster_options}} are required")
 		}
 
 		if ghac.RedisTokenDB != nil && ghac.RedisTokenDB.EncryptKey != "" && len(ghac.RedisTokenDB.EncryptKey) != 32 {
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -200,6 +200,10 @@ func validate(c *Config) error {
 			return errors.New("github_auth.{client_id,client_secret,redis_token_db{url}} are required")
 		}
 
+		if ghac.RedisTokenDB != nil && ghac.RedisTokenDB.EncryptKey != "" && len(ghac.RedisTokenDB.EncryptKey) != 32 {
+			return errors.New("github_auth.{redis_token_db{encrypt_key}} must be exactly 32 bytes long")
+		}
+
 		if ghac.HTTPTimeout <= 0 {
 			ghac.HTTPTimeout = time.Duration(10 * time.Second)
 		}
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -200,10 +200,6 @@ func validate(c *Config) error {
 			return errors.New("github_auth.{client_id,client_secret,redis_token_db.{node_options,cluster_options}} are required")
 		}
 
-		if ghac.RedisTokenDB != nil && ghac.RedisTokenDB.EncryptKey != "" && len(ghac.RedisTokenDB.EncryptKey) != 32 {
-			return errors.New("github_auth.{redis_token_db{encrypt_key}} must be exactly 32 bytes long")
-		}
-
 		if ghac.HTTPTimeout <= 0 {
 			ghac.HTTPTimeout = time.Duration(10 * time.Second)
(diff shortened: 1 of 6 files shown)


karmi:
@rojer , also when you have time, can you have a look at this one? I know it's an invasive/noisy patch, but when the app is running distributed, Redis is arguably an attractive option to store tokens...

fxdgear:
ping @rojer or anyone have time to review this patch?

rojer:
wow, it's been a while. this needs to be rebased to master.

fxdgear:
@karmi can you rebase?

flaper87:
> @karmi can you rebase?

I'll do it!

flaper87:
Ok, It is rebased now :)

rojer:
thanks @flaper87 
1. i am not a fan of having two similar config keys, url and urls. i'd leave just urls.
2. not a huge fan of the encryption part. why is it necessary to encrypt stuff in the redis database?

flaper87:
Thank you for your comment. I'll address #1, not a fan of it myself either. I limited myself to just rebasing the patch. in the previous push.

As for #2, I've asked @karmi to please shed some light as I'm just starting to familiarize myself with this PR. Maybe @dliappis can also answer the question.

dliappis:
@flaper87 The motivation for 2. -- memory maybe vague -- was:

1. perhaps this auth is configured against a Redis service administered by a different organization to the one managing the Docker Registry or a Cloud based service; encryption provides some protection.

2. typically lots of Redis installations don't have auth enabled.

Note that encryption is optional.

karmi:
Hello, @rojer! What a trip down the memory lane! :)

This is actually the first significant piece of Go I've written, and chances are, I would write it quite differently these days. However, let me address the questions you bring up.

> i am not a fan of having two similar config keys, url and urls. i'd leave just urls.

As said, it's been quite some time ago, but what I remember is that the `redis` package uses two different constructors when Redis is running a standalone mode, and when in the cluster mode. We found it the hard way, when the code was working fine for us during development, but failed in a staging environment in AWS — because there, Redis has been running in the cluster mode.

Of course, it would be better to just use `urls`, but the disctinction between the `url` and `urls` configuration option was the easiest, and least invasive option we saw at the time. 

> not a huge fan of the encryption part. why is it necessary to encrypt stuff in the redis database?

My memory is even more hazy on this point, but I guess our thinking was something along the lines of "better safe then sorry". As you can imagine, leaking the token database could have catastrophic effects for a private registry, and also, since the token can be considered a piece of personal data under GDPR regulation, we went with the general advice to encrypt such data.

rojer:
would using non-cluster client if len(urls) == 1 be a reasonable solution? or are clusters of just one node a widely used thing?

regarding encryption, i'd rather we not do that. in the current form it's incomplete - it stores secrets (symmetric encrytion key) in configuration file and doesn't provide for rotation mechanism (what if you need to change the key?). imo, encryption of backend data is better left to external means - use a properly secure redis instance or deploy an encrypting proxy of some sort. it doesn't belong in docker_auth.

karmi:
Right, using ` len(urls) == 1` as the condition would be optimal, but if memory serves, unfortunately on AWS ElastiCache (the Redis service we have been using), there's indeed only a single URL to connect to. Maybe @fxdgear or @flaper87 have some ideas here? It looks like a good way to get familiar with the codebase.

I do agree that it would be best to leave the encryption to an external service. For us, it was really only a nice way how to make the implementation more robust security-wise — if the Redis instance would be compromised, the tokens are useless, since they are encrypted.

flaper87:
> Right, using ` len(urls) == 1` as the condition would be optimal, but if memory serves, unfortunately on AWS ElastiCache (the Redis service we have been using), there's indeed only a single URL to connect to. Maybe @fxdgear or @flaper87 have some ideas here? It looks like a good way to get familiar with the codebase.
> 
> I do agree that it would be best to leave the encryption to an external service. For us, it was really only a nice way how to make the implementation more robust security-wise — if the Redis instance would be compromised, the tokens are useless, since they are encrypted.


I took a stab at this and replaced the `Url/Urls` options in favor of the `go-redis` `Options` and `ClusterOptions` types. This should reduce the customization in docker_auth and it'll make it easier for users to configure their redis options.

@rojer Would you be open to considering keeping the `encryption` part of the code after @karmi and @dliappis explanation on why it was added?

flaper87:
> @rojer Would you be open to considering keeping the `encryption` part of the code after @karmi and @dliappis explanation on why it was added?

nvm, I just re-read your previous comment and your concerns about the lack of rotation for the key.  I'll get rid of that part for now so that we don't block this PR on it.

The more I think about it, the more I'm convinced it'd be nice for `docker_auth` to provide support for encryption, either through an external service/implementation or an internal one. This certainly doesn't belong to the DB code.

flaper87:
@rojer I've removed the encryption support for the redis token DB. I think this PR is ready for a review now. :)

Thanks a bunch

auth_server/server/config.go
```diff
@@ -188,13 +188,18 @@ func validate(c *Config) error {
...
-		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.TokenDB == "" && ghac.GCSTokenDB == nil) {
+		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.TokenDB == "" && (ghac.GCSTokenDB == nil && ghac.RedisTokenDB == nil)) {
 			return errors.New("github_auth.{client_id,client_secret,token_db} are required")
 		}
 
 		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.GCSTokenDB != nil && (ghac.GCSTokenDB.Bucket == "" || ghac.GCSTokenDB.ClientSecretFile == "")) {
 			return errors.New("github_auth.{client_id,client_secret,gcs_token_db{bucket,client_secret_file}} are required")
 		}
+
+		if ghac.ClientId == "" || ghac.ClientSecret == "" || (ghac.RedisTokenDB != nil && ghac.RedisTokenDB.NodeOptions.Addr == "" && len(ghac.RedisTokenDB.ClusterOptions.Addrs) < 1) {
```
flaper87: I left this change here but we should be able to do away with it. `go-redis` has a default value for `Addr` and `Addrs`. I kept it mainly because handling the configuration options will be weird if we don't set one of these explicitly:

For example:

```
redis_token_db:
    node_options:
    cluster_options:
```

The above would be a valid config but we won't know if we should create a single node connection or a cluster one. We can default to the single node one but it will be confusing to users.

rojer:
there are currently no other options in these, other than address(es).
can we pull them up a level and just have `redis_addr` and `redis_cluster_addrs`?

flaper87:
> there are currently no other options in these, other than address(es).
> can we pull them up a level and just have `redis_addr` and `redis_cluster_addrs`?

I originally did something like this but then changed. 

To be honest, I would prefer to not do this. There are other options that could be set (like the DB index / TLS to use in the case of the [Client](https://godoc.org/github.com/go-redis/redis#Options)  or the PoolSize/TLS in the case of the [ClusterOptions](https://godoc.org/github.com/go-redis/redis#ClusterOptions) )

I think it's important to not prevent users from setting the options for the Redis connection.

rojer:
ok, fair enough.

thanks everyone for contributing!

PR #225, opened 2018-02-19
Add env variable support for configuration

umurgdk:
This commit also makes possible to use json files as configuration
files via `viper` configuration library.

This PR contains breaking changes on configuration to make
configuration overridable by the environment variables. To make it
support nested struct overrides struct tags are renamed to not
contain any underscore in their names. **This is going to break old
configuration files.**

Because of viper's limitations of detecting new map keys from env
variables, a little procedure goes trough prefixed env variables and
override viper settings by the parsed value of them. If the
configuration file is in yaml format env variables can contain correct
yaml strings and they will be parsed correctly. This behaviour is same
for json files too.

PS: viper doesn't use file format parsers for unmarshalling the structures.
It parses files as `interface{}` and the using `mapstructure` library to
unmarshall structs from maps. This is why `yaml` struct tags converted to
`mapstructure` ones.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/225)

### DIFF
--- a/auth_server/authn/ext_auth.go
+++ b/auth_server/authn/ext_auth.go
@@ -27,8 +27,8 @@ import (
 )
 
 type ExtAuthConfig struct {
-	Command string   `yaml:"command"`
-	Args    []string `yaml:"args"`
+	Command string   `mapstructure:"command"`
+	Args    []string `mapstructure:"args"`
 }
 
 type ExtAuthStatus int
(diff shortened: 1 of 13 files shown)


umurgdk:
I will update the readme after the code review.

techknowlogick:
Thanks for this PR :)

The codebase has evolved since this PR was made, and so I will close it. If you are open to re-submitting it and resolving the conflicts, I think it would be a benefit to the project.

ewassef:
I would like to reopen this.. i have updated the PR @umurgdk @techknowlogick

Issue #231, opened 2018-05-03
TLS 1.2?


Spiltpickle:
Would it be possible to be able to define which TLS version to use in the config?

This is to fulfill PCI requirements of not using TLS 1.0, etc...

adamdecaf:
https://github.com/cesanta/docker_auth/pull/232 should let us use TLS 1.2

PR #232, opened 2018-05-31
drop old hardcoded configs around TLS 1.x

adamdecaf:
Modern Go versions (1.9 and 1.10) as of this commit are much better
about cipher suite selection and the ssl/tls protocols used. In fact,
SSLv3 needs to be explicitly enabled now.

Fixes: https://github.com/cesanta/docker_auth/issues/231

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/232)


rojer:
thanks!

PR #235, opened 2018-06-25
authz: printf string value in log, not pointer

adamdecaf:
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/235)


rojer:
problem is, it can be NULL and then things will blow up.
tbh, there's probably not a good reason Comment is a pointer to string rather than string, you can try changing it.

techknowlogick:
Thank you for your PR. This has been resolved in latest commit, and this PR can now be closed.

PR #239, opened 2018-07-26
Add circle CI

kofalt:
Added a quick circle CI config. Free service, runs fast, [looks like this](https://circleci.com/gh/kofalt/docker_auth/4).
You'll have to enable it through their app after merging.

It'd be pretty easy to add a gofmt check, or to have it save the binary after compiling so people can download test builds. The build fails right now because I invoke `make` without `Version` or `BuildId`, what would you like for me to set there?

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/239)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,22 @@
+version: 2
+
+jobs:
+  build:
+
+    working_directory: /go/src/github.com/cesanta/docker_auth
+
+    docker:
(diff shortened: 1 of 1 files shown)


techknowlogick:
@kofalt Thanks for this PR. This will be closed in favour of moving to Github actions (as it is one less service to maintain)

PR #228, opened 2018-04-12
MAP LDAP account attributes to labels such as groups

mrueg:
Since the original PR went stale, I
* Cherry-picked the commits from https://github.com/cesanta/docker_auth/pull/167 so they apply against master again. 
* Added example ACL as requested

@rojer @alonbl Please have another look. Also @kcd83 fyi :)

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/docker_auth/228)

PR #247, opened 2019-01-29
Add support for lowercases DN elements

pinepain:
A follow up to https://github.com/cesanta/docker_auth/pull/228#issuecomment-389545187 comment:

> FreeIPA seems to have types of DN elements lowercased

In this PR we transform DN attributes to upper case (as in most cases they are already uppercased, so it looks like it's more optimal then lowercasing) and compare to `"CN"`.


pinepain:
@jaen fyi

jaen:
@pinepain oh, that's awfully nice of you ;' )

The people who inherited the infrastructure will be very happy to know they can switch to the upstream image when that is in.

rojer:
merged, thanks!

PR #251, opened 2019-05-15
return 401 when there is no user or password

duyanghao:
Signed-off-by: duyanghao <1294057873@qq.com>


rojer:
thank you!

rojer:
unfortunately this PR breaks an important use case, so i reverted it (https://github.com/cesanta/docker_auth/commit/f1a115f5e2a7f6ce2d4fc07b833bf2c704fdf3e8).

PR #259, opened 2019-09-05
feat: extended TLS configuration

trutx:
This PR extends the server's TLS configuration. While the default TLS settings should be enough for most use cases, if you get a bit paranoid about security you'd probably want to tweak things a bit. This PR allows that.

The motivation in my case is passing the FedRAMP certification, which required some TLS vulnerabilities to be addressed. Here's a FedRAMP compliant (as of Sep 5th, 2019) configuration example:

```yaml
server:
  addr: :5001
  certificate: /path/to/server.pem
  key: /path/to/server.key
  hsts: true
  tls_min_version: 0x0303
  tls_curve_preferences:
    - 25
    - 24
    - 23
  tls_cipher_suites:
    - 0x009c
    - 0x009d
    - 0xc02f
    - 0xc02b
    - 0xc030
    - 0xc02c
    - 0xcca8
    - 0xcca9
```
If you're curious I used [this tool](https://github.com/drwetter/testssl.sh) to test both the current server and the patched one with the settings above, and the differences are there.

### DIFF
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -113,7 +113,7 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
 			glog.Exitf("Failed to load certificate and key: both were not provided")
 		}
 		glog.Infof("Cert file: %s", c.Server.CertFile)
-		glog.Infof("Key file: %s", c.Server.KeyFile)
+		glog.Infof("Key file : %s", c.Server.KeyFile)
 		tlsConfig.Certificates = make([]tls.Certificate, 1)
 		tlsConfig.Certificates[0], err = tls.LoadX509KeyPair(c.Server.CertFile, c.Server.KeyFile)
 		if err != nil {
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -53,13 +53,28 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
 	tlsConfig := &tls.Config{
 		PreferServerCipherSuites: true,
 	}
+	if c.Server.HSTS {
+		glog.Info("HTTP Strict Transport Security enabled")
+	}
+	if c.Server.TLSMinVersion > 0 {
+		tlsConfig.MinVersion = c.Server.TLSMinVersion
+		glog.Infof("TLS MinVersion: %#04x", tlsConfig.MinVersion)
+	}
+	if len(c.Server.TLSCurvePreferences) > 0 {
+		tlsConfig.CurvePreferences = c.Server.TLSCurvePreferences
+		glog.Infof("TLS CurvePreferences: %d", tlsConfig.CurvePreferences)
+	}
+	if len(c.Server.TLSCipherSuites) > 0 {
+		tlsConfig.CipherSuites = c.Server.TLSCipherSuites
+		glog.Infof("TLS CipherSuites: %#04x", tlsConfig.CipherSuites)
+	}
 	if c.Server.CertFile != "" || c.Server.KeyFile != "" {
 		// Check for partial configuration.
 		if c.Server.CertFile == "" || c.Server.KeyFile == "" {
 			glog.Exitf("Failed to load certificate and key: both were not provided")
 		}
 		glog.Infof("Cert file: %s", c.Server.CertFile)
-		glog.Infof("Key file : %s", c.Server.KeyFile)
+		glog.Infof("Key file: %s", c.Server.KeyFile)
 		tlsConfig.Certificates = make([]tls.Certificate, 1)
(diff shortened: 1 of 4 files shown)


rojer:
no objections in general, but for readability, can we use string names for the ciphers and curve names instead?

rojer:
`TLS_RSA_WITH_AES_128_GCM_SHA256` is much more readable than 0x009c
if there is no existing method to look up cipher suite number by name, you could use reflect to look up symbol by name in the tls package.

trutx:
I'm not sure I understand what you mean by using `reflect` to lookup constants from a remote package. If it can be done I'd definitely love to learn how.

My goal was to keep the code simple and to avoid hardcoding (I guess that was the goal in https://github.com/cesanta/docker_auth/pull/232/files too). I see no way around either a hardcoded map to "translate" user provided strings to actual `tls.CipherSuite`s or some sort of pre-build `go generate` that does a similar thing, but again I'd love to learn a new solution. I totally agree that the actual ciphersuite name is much more readable and user friendly than a hex value.

rojer:
ok, scratch that, you cannot use reflect to lookup package constants.

so, how about we support string representation for [existing list of ciphers](https://golang.org/pkg/crypto/tls/#pkg-constants) as `map[string]uint16` but also allow numeric values.

that way:

```yaml
tls_cipher_suites:
  - TLS_RSA_WITH_RC4_128_SHA
 - 0x0005
```

will both work. all you need to do is make that list of strings, look up value in the hard-coded map and then fall back to `strconv.Atoi`.

same applies to curve names.

trutx:
That should do it. It was slightly trickier than `strconv.Atoi` because we're dealing with `uint16` here. Now you can pass in either a string or an hex value to configure the TLS min version, the TLS Curves and the TLS CipherSuites.

auth_server/main.go
```diff
@@ -53,13 +54,66 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
...
+		glog.Infof("TLS CipherSuites: %s", c.Server.TLSCipherSuites)
+	}
 	if c.Server.CertFile != "" || c.Server.KeyFile != "" {
 		// Check for partial configuration.
 		if c.Server.CertFile == "" || c.Server.KeyFile == "" {
 			glog.Exitf("Failed to load certificate and key: both were not provided")
 		}
 		glog.Infof("Cert file: %s", c.Server.CertFile)
-		glog.Infof("Key file : %s", c.Server.KeyFile)
+		glog.Infof("Key file: %s", c.Server.KeyFile)
```
rojer: keep this space, it makes the lines in the output aligned and more readable

auth_server/server/config.go
```diff
@@ -78,14 +82,75 @@ type TokenConfig struct {
...
+	// TLS_FALLBACK_SCSV isn't a standard cipher suite but an indicator
+	// that the client is doing version fallback. See RFC 7507.
+	"TLS_FALLBACK_SCSV": tls.TLS_FALLBACK_SCSV,
+}
+
+// TLSVersionValues maps Version names as strings to the actual values in the
+// crypto/tls package
+// Taken from https://golang.org/pkg/crypto/tls/#pkg-constants
+var TLSVersionValues = map[string]uint16{
+	"VersionTLS10": tls.VersionTLS10,
```
rojer: drop the Version prefix, it's more readable

auth_server/server/config.go
```diff
@@ -78,14 +82,75 @@ type TokenConfig struct {
...
+	// Deprecated: SSLv3 is cryptographically broken, and will be
+	// removed in Go 1.14. See golang.org/issue/32716.
+	"VersionSSL30": tls.VersionSSL30,
+}
+
+// TLSCurveIDValues maps CurveID names as strings to the actual values in the
+// crypto/tls package
+// Taken from https://golang.org/pkg/crypto/tls/#CurveID
+var TLSCurveIDValues = map[string]tls.CurveID{
+	"CurveP256": tls.CurveP256,
```
rojer: drop the Curve prefix, don't need it

auth_server/main.go
```diff
@@ -53,13 +54,66 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
 	tlsConfig := &tls.Config{
 		PreferServerCipherSuites: true,
 	}
+	if c.Server.HSTS {
+		glog.Info("HTTP Strict Transport Security enabled")
+	}
+	if c.Server.TLSMinVersion != "" {
+		var value uint16
```
rojer: there's a lot of repetition here, please move to a helper function

trutx:
Some validation logic could be added in `config.go` to check that all the passed values exist in the value maps, though this would mean doing the conversion dance twice -- one in `server.validate()` and another one in `ServeOnce()`. Not sure if it's worth it.

rojer:
merged, thank you!

PR #261, opened 2019-09-19
add helm chart for docker_auth

duyanghao:
Add helm chart of docker_auth that supports High-Available.

Test successfully with following registry config.yaml:
```
auth:
  token:
    autoredirect: false
    realm: https://docker-auth.domain/auth
    service: token-service
    issuer: Acme auth server
    rootcertbundle: /data/duyanghao/registry-test/server.pem
```
Signed-off-by: duyanghao <1294057873@qq.com>


rojer:
ok, i have no idea what this is and it's kinda big.
can you explain a little bit? does it absolutely have to be in the repo?

duyanghao:
> ok, i have no idea what this is and it's kinda big.
> can you explain a little bit? does it absolutely have to be in the repo?

@rojer 
>>>Helm uses a packaging format called charts. A chart is a collection of files that describe a related set of Kubernetes resources. A single chart might be used to deploy something simple, like a memcached pod, or something complex, like a full web app stack with HTTP servers, databases, caches, and so on.
This is only a helm chart of docker_auth, which does not affect the project itself. Adding helm chart is to make it easier for docker_auth to be deployed into K8s.

auth_server/chart/docker-auth/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+# Declare variables to be passed into your templates.
+
+replicaCount: 3
+
+image:
+  repository: cesanta/docker_auth
+  tag: 1.4.0
+  pullPolicy: IfNotPresent
+
+configmap:
```
carsonoid: I would say that using the key `configmap` here is confusing. Because some of this data ends up in a Secret and not a ConfigMap. People looking at the values file and not the templates may wonder if you are going to put the tls files into an insecure location. The simple solution may just be to use the word `config` instead which wouldn't have that connotation.

auth_server/chart/docker-auth/templates/configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+data:
+  auth_config.yml: |
+    server:
+      addr: ":{{ .Values.service.targetPort }}"
+    token:
+      issuer: "{{ .Values.configmap.data.token.issuer }}"  # Must match issuer in the Registry config.
+      expiration: {{ .Values.configmap.data.token.expiration }}
+      certificate: "/config/certs/server.pem"
+      key: "/config/certs/server.key"
+    users:
```
carsonoid: Doing users this way means that the chart only supports two uses out of the box. The config files are already yaml. So I would recommend that you just use `{{ .Values.config.users | toYaml | nindent 6 }}` to inject the entire users section as embedded yaml.

auth_server/chart/docker-auth/templates/configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+      # Password is specified as a BCrypt hash. Use `htpasswd -nB USERNAME` to generate.
+      "{{ .Values.configmap.data.users.admin.username }}":
+        password: "{{ .Values.configmap.data.users.admin.password }}"
+      "{{ .Values.configmap.data.users.test.username }}":
+        password: "{{ .Values.configmap.data.users.test.password }}"
+    acl:
```
carsonoid: You should update and implement the same pattern as suggested for users.

auth_server/chart/docker-auth/values.yaml
duyanghao: Thank you for pointing that. I have split `configmap` into two parts: `configmap` and `secret`, then connotation will disappear.

auth_server/chart/docker-auth/templates/configmap.yaml
duyanghao: Good advice, Done for that.

auth_server/chart/docker-auth/templates/configmap.yaml
duyanghao: Done.

auth_server/chart/docker-auth/README.md
```diff
@@ -0,0 +1,89 @@
...
+    issuer: Acme auth server
+    rootcertbundle: /path/to/server.pem
+```
+
+### Uninstallation
+
+To uninstall/delete the `my-release` deployment:
+
+```bash
+# helm delete --purge docker-auth
```
carsonoid: Get rid of the `#` here or replace it with `$`, match whatever you choose for the other change above.

auth_server/chart/docker-auth/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Helm 2.8.0+
+
+## Installation
+
+### Install the chart
+
+Install the docker-auth helm chart with a release name `my-release`:
+
+```bash
+# helm install --name=docker-auth docker-auth
```
carsonoid: Get rid of the `#` here. I recognize that you are trying to visually distinguish the execution of the command and the results. But having it rendered as a comment is a bit confusing. You should just use two different blocks to make that distinction.  Or at a minimum, use the semi-recognized pattern of using a `$` to indicate the command.

```
$ helm install --name=docker-auth docker-auth
OUTPUT HERE
```

or

```
helm install --name=docker-auth docker-auth
```
```
OUTPUT HERE
```

auth_server/chart/docker-auth/README.md
duyanghao: Thank you for pointing that, Done for it.

auth_server/chart/docker-auth/README.md
duyanghao: Done.

duyanghao:
> I have a few last nitpicks. The other big change I would suggest is to move the whole charts directory to the root of the repo since that makes more sense to me.
> 
> Once those last things get done I would say this looks good to go and it's up to @rojer to decide if this belongs here or in the helm repos.
> 
> By the way, thanks for all the work on this @duyanghao! I recognize that you could have just as easily kept this private and not contributed back.

@carsonoid I am glad I could contribute in some way and thanks for the advice about this pr.

carsonoid:
@rojer LGTM

rojer:
@carsonoid thanks for the review! i'm going to merge this withot understanding in a hope that it's useful :)

pfisterer:
@duyanghao Thank your for contributing the helm charts. Actually, I'd really love to see your work being available on helm hub. 

I also wanted to contribute a new chart (https://github.com/helm/charts/pull/16623) but they don't accept new charts (cf. https://github.com/helm/charts/pull/16623#issuecomment-526899438) and encourage you to host your own helm repo.

I did this for my chart here: https://github.com/pfisterer/apache-knox-helm/

Are you interested in developing/improving this chart together here on github? I can create a repo with the current charts and add you as contributor (or vice versa).

pfisterer:
I successfully managed to run docker_auth and docker-registry with a modified version of the great helm charts contributed by @duyanghao. 

I have uploaded them here: https://github.com/pfisterer/docker-auth-helm.

rojer:
@pfisterer if you made changes that you think make sense to be upstreamed, please open a pull request.

pfisterer:
@rojer: done. Sorry, but this is a quite large PR. Cf. #263

PR #267, opened 2019-11-28
Existing secret support for helm chart

schegi:
Adding support for using an existing k8s secret providing server certificate and key. This is usefull if you generate our secrets using e.g. the cert-manager.


chart/docker-auth/README.md
```diff
@@ -76,8 +76,11 @@ The following table lists the configurable parameters of the docker-auth chart a
...
+| `secret.data.server.key`                                                     | Content of server.key  (mutually exclusive with secretName, keyName, certificateName)                                                                                                                                                                                                                                                                                                                        |                           |
+| `secret.secretName`                                                          | The name of the secret containing server key and certificate (mutually exclusive with secret.data.server.key/certificate)         | |
+| `secret.certificateName`                                                     | The name of the server certificate file (mutually exclusive with secret.data.server.key/certificate)                               | |
```
carsonoid: Cert-Manager will always use the key `tls.crt` so you should set that as the default for the certificateName

chart/docker-auth/README.md
```diff
@@ -76,8 +76,11 @@ The following table lists the configurable parameters of the docker-auth chart a
...
+| `secret.keyName`                                                             | The name of the server key file (mutually exclusive with secret.data.server.key/certificate)                                       | |
```
carsonoid: Cert-Manager will always use the key `tls.key` so you should set that as the default for the keyName

chart/docker-auth/values.yaml
```diff
@@ -18,6 +18,12 @@ secret:
       key: |+
         -----BEGIN RSA PRIVATE KEY-----
         -----END RSA PRIVATE KEY-----
+# For reusing an existing secret (e.g. generated by cert-manager), define secretName, certificateFile and keyFileName
+# These settings are mutually exclusive with the values provided in secret.data. Once secretName is et the secret
+# generated with the values above will be not be used in the deployment.
+#  secretName:
+#  certificateFileName:
```
carsonoid: Update with default. Also this is called `certificateName` in the readme.

chart/docker-auth/values.yaml
```diff
@@ -18,6 +18,12 @@ secret:
...
+#  keyFileName:
```
carsonoid: Update with default. Also this is called `keyName` in the readme.

chart/docker-auth/templates/configmap.yaml
```diff
@@ -9,8 +9,16 @@ data:
     token:
       issuer: "{{ .Values.configmap.data.token.issuer }}"  # Must match issuer in the Registry config.
       expiration: {{ .Values.configmap.data.token.expiration }}
+{{- if .Values.secret.certificateName }}
```
carsonoid: Since `certificateName` will have a default, check for `.Values.secret.secretName` instead.

chart/docker-auth/templates/configmap.yaml
```diff
@@ -9,8 +9,16 @@ data:
...
+      certificate: "/config/certs/{{ .Values.secret.certificateName}}"
+{{- else }}
       certificate: "/config/certs/server.pem"
+{{- end }}
+{{- if .Values.secret.keyName }}
```
carsonoid: Since `keyName` will have a default, check for `.Values.secret.secretName` instead.

carsonoid: You can also merge this if statement with the one above because they will be checking for the same thing.

schegi:
Did all you requestsed.

rojer:
thanks for review, @carsonoid , merging

PR #270, opened 2020-01-31
Send the scope class as part of the token

### LINKED ISSUE #269: docker plugin push fails with insufficent scope
I'm trying to figure out why I can't push a plugin to my registry with auth provided by this project.

when I push an image everything works fine. When I push a plugin I get an error insufficient scope. 

I tried reading through the auth section of this project but it doesn't seem to really call out any particular scope types inparticular so I figure it should be pretty generic, as long as you're properly conforming to the grammar. (which you reference that you are). 

So, I was wondering if there's any flags I need to change or anything in particular I should be made aware of when using docker_auth for pushing plugins into my repository. 

Thanks

flaper87:
As it is right now, it is not possible to authenticate plugin requests using `docker-auth`. The docker distribution registry expects[0] a scope with `Type` repository and `docker-auth` is sending `repository(plugin)` as that's what it receives in the request.

Instead of sending `repository(plugin)` we should be sending `repository` and setting the scope class to `plugin`. The default class is `image`, which is why this works fine right now.

I had trouble finding a detailed documentation about how the docker registry uses `class` but here's a test:
 
```curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository(plugin):rchicoli/docker-log-elasticsearch:pull,push"```

The above command produces a token that contains the following access request (you can decode the token using https://jwt.io/ [thanks for the hint @fxdgear) ]:

```
...
"access": [
    {
      "type": "repository",
      "class": "plugin",
      "name": "rchicoli/docker-log-elasticsearch",
      "actions": [
        "pull"
      ]
    }
  ],
...
```

Fixes #269

### DIFF
--- a/auth_server/server/server.go
+++ b/auth_server/server/server.go
@@ -38,6 +38,7 @@ import (
 
 var (
 	hostPortRegex = regexp.MustCompile(`\[?(.+?)\]?:\d+$`)
+	scopeRegex    = regexp.MustCompile(`([a-z0-9]+)(\([a-z0-9]+\))?`)
 )
 
 type AuthServer struct {
@@ -138,6 +139,7 @@ type authRequest struct {
 
 type authScope struct {
 	Type    string
+	Class   string
 	Name    string
 	Actions []string
 }
@@ -158,6 +160,22 @@ func parseRemoteAddr(ra string) net.IP {
 	}
 	res := net.ParseIP(ra)
 	return res
+}
+
+func parseScope(scope string) (string, string, error) {
+	parts := scopeRegex.FindStringSubmatch(scope)
+	if parts == nil {
+		return "", "", fmt.Errorf("malformed scope request")
+	}
+
+	switch len(parts) {
+	case 3:
+		return parts[1], "", nil
+	case 4:
+		return parts[1], parts[3], nil
+	default:
+		return "", "", fmt.Errorf("malformed scope request")
+	}
 }
 
 func (as *AuthServer) ParseRequest(req *http.Request) (*authRequest, error) {
@@ -212,16 +230,24 @@ func (as *AuthServer) ParseRequest(req *http.Request) (*authRequest, error) {
 		for _, scopeStr := range req.Form["scope"] {
 			parts := strings.Split(scopeStr, ":")
 			var scope authScope
+
+			scopeType, scopeClass, err := parseScope(parts[0])
+			if err != nil {
+				return nil, err
+			}
+
 			switch len(parts) {
 			case 3:
 				scope = authScope{
-					Type:    parts[0],
+					Type:    scopeType,
+					Class:   scopeClass,
 					Name:    parts[1],
 					Actions: strings.Split(parts[2], ","),
 				}
 			case 4:
 				scope = authScope{
-					Type:    parts[0],
+					Type:    scopeType,
+					Class:   scopeClass,
 					Name:    parts[1] + ":" + parts[2],
 					Actions: strings.Split(parts[3], ","),
 				}

### CODE REVIEW

File: auth_server/server/server.go [hunk 1]
```diff
@@ -212,16 +213,29 @@ func (as *AuthServer) ParseRequest(req *http.Request) (*authRequest, error) {
 		for _, scopeStr := range req.Form["scope"] {
 			parts := strings.Split(scopeStr, ":")
 			var scope authScope
+
+			scopeType := parts[0]
+			var scopeClass string
+			if strings.HasPrefix(scopeType, "repository") {
```
rojer: instead of prefix-suffix matching, can't we just compare to "repository(plugin)"?

flaper87: According to [this document](https://docs.docker.com/registry/spec/auth/scope/) there's a third scope that could be sent. This is why I went with matching tring parts.

rojer: cool, that's a good document.
the way i read it, scope consists of resource type and optional class, i.e. `TYPE` or `TYPE(CLASS)`.
can you write a helper function that parses this format and returns resourceType, resourceClass and error (if rformat is not recognized) ?

flaper87: yes, let's do that!

flaper87: Ok, this is done! Thanks for the review :)

auth_server/server/server.go
```diff
@@ -38,6 +38,7 @@ import (
 
 var (
 	hostPortRegex = regexp.MustCompile(`\[?(.+?)\]?:\d+$`)
+	scopeRegex    = regexp.MustCompile("(\\w+)($|\\((\\w+)\\))")
```
rojer: use backtick quoting to avoid "\\" stuttering.
use character class exactly as stated in the doc - `[a-z0-9]+`

flaper87: a-ha! Good catch. Fixed it! :)

auth_server/server/server.go
```diff
@@ -38,6 +38,7 @@ import (
 
 var (
 	hostPortRegex = regexp.MustCompile(`\[?(.+?)\]?:\d+$`)
+	scopeRegex    = regexp.MustCompile(`([a-z0-9]+)($|\(([a-z0-9]+)\))`)
```
rojer: `([a-z0-9]+)(\([a-z0-9]+\))?`

flaper87: Won't it produce the same result? Why do you prefer this other regex?

rojer: yes, it better conveys the optional nature of type.

rojer:
ok, can you please verify it works? with plugin and non=-plugin requests.

flaper87:
I did already :)

Plugin push:

```
~/…/docker_auth/auth_server $ docker plugin push localhost:5000/mine/my-plugin:0.0.1
The push refers to repository [localhost:5000/mine/my-plugin]
d3adcb574d5b: Pushed
0.0.1: digest: sha256:3c62fd743e3d8f1eed511969dc16d74c043cf9b24fa3522f92d19a388a6deae7 size: 519
```

Repository push

```
~/…/docker_auth/auth_server $ docker push localhost:5000/mine/docker_auth
The push refers to repository [localhost:5000/mine/docker_auth]
9db3677022cc: Layer already exists
0537e79a9821: Layer already exists
cd7100a72410: Layer already exists
latest: digest: sha256:3f5e60df82833e39e5179e02400b302aa9290790539233aa548cd6bcfae60f78 size: 949
```

docker auth logs:

```
I0131 16:36:30.080223 1967670 server.go:379] New token for {flaper87:***@[::1]:57978 [{repository  mine/my-plugin [pull push]}]} map[teams:[REDACTED]]: {"iss":"Acme auth server","sub":"flaper87","aud":"Docker registry","exp":1580489490,"nbf":1580488580,"iat":1580488590,"jti":"6276831680445898189","access":[{"type":"repository","name":"mine/my-plugin","actions":["pull","push"]}]}
I0131 16:36:35.972957 1967670 server.go:379] New token for {flaper87:***@[::1]:58000 [{repository  mine/docker_auth [pull push]}]} map[teams:[REDACTED]]: {"iss":"Acme auth server","sub":"flaper87","aud":"Docker registry","exp":1580489495,"nbf":1580488585,"iat":1580488595,"jti":"1270047298924844124","access":[{"type":"repository","name":"mine/docker_auth","actions":["pull","push"]}]}
```

rojer:
thank you!

fxdgear:
Yay!!!! Thank you both.

flaper87:
> thank you!

thank you for the review and for maintaining this project

rojer:
> > thank you!
> 
> thank you for the review and for maintaining this project

yeah, at this point i'm not even running the thing anymore :)

flaper87:
We (elastic) are running it and we'd be happy to help maintaining the project, if you'd have us. 😊🤗

rojer:
your contributions are welcome! if you want stuff to do - check out issues.
one big area that needs improvement is testing, unit and integration.

Issue #272, opened 2020-02-11
Authorization Service should also return the "token"-field


dennismeissel:
There are old docker-registry clients, that use the `token` field instead of `access_token`.
Only adding a `token` field with the same content will solve the compability problem with these clients.
More information is available [here](https://docs.docker.com/registry/spec/auth/token/#token-response-fields)

dennismeissel:
Fixed in #273

PR #273, opened 2020-02-11
Returning the "token"-field

dennismeissel:
This change is necessary to support older clients. #272


rojer:
please add a comment in the code why we return both

dennismeissel:
Done

rojer:
thanks!

PR #274, opened 2020-03-06
Fix scope parsing

jsravn:
Per
https://docs.docker.com/registry/spec/auth/scope/#resource-scope-grammar
and https://docs.docker.com/registry/spec/auth/oauth/
multiple scopes are separated by a single space within a single `scope`
parameter.

This fixes compatibility with buildkit which currently generates a bad request when pushing:

``` sh
W0306 15:29:11.332525      31 server.go:424] Bad request: invalid scope: "repository:my-registry.com/my-repo:pull repository:my-registry.com/my-repo:pull,push"
```

### DIFF
--- a/auth_server/server/server.go
+++ b/auth_server/server/server.go
@@ -226,37 +226,38 @@ func (as *AuthServer) ParseRequest(req *http.Request) (*authRequest, error) {
 		return nil, fmt.Errorf("invalid form value")
 	}
 	// https://github.com/docker/distribution/blob/1b9ab303a477ded9bdd3fc97e9119fa8f9e58fca/docs/spec/auth/scope.md#resource-scope-grammar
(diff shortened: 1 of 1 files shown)


rojer:
it also says, multiple "scope" headers can be used in get requests but your change removes handling of multiple headers.
since we don't really distinguish between get and post, i'd rather we preserve it.
so, iterate over all scope headers and additionally split each of them to multiple scopes.

jsravn:
@rojer good catch, I missed that. It's confusing that GET and POST have different semantics for the header. I will fix it.

jsravn:
Okay, it should now support both GET and POST approaches.

rojer:
thank you!

rojer:
pushed to :latest

PR #282, opened 2020-06-03
Fix Docker Hub link in README.md

Strayer:
Really just a tiny documentation fix, but the Docker Hub link in the README.md currently points to a wrong URL, leading the reader to an empty page.


rojer:
thank you!

PR #285, opened 2020-07-13
Move to Python3

JamesWrigley:
This fixes the `build-release` target. The builder image is based on alpine, which as of 3.12 no longer includes `py2-pip`.


rojer:
thank you!

PR #293, opened 2020-10-12
Upgrade go deps

techknowlogick:
* Use go-bindata that receives updates
* Switch make build-release to use go1.15
* Update alpine pkg install to use updated name
* run go get -u
* move go generate for go-bindata to directory that contains files to embed


rojer:
thank you!

PR #296, opened 2020-12-28
add gitea authn

cgroschupp:


auth_server/Makefile
```diff
@@ -22,7 +22,7 @@ generate:
 	  github.com/cesanta/docker_auth/auth_server/server/...
 
 build:
-	CGO_ENABLED=0 go build -v --ldflags=--s
+	CGO_ENABLED=0 go build -v --ldflags=--s -pkgdir /tmp/go
```
techknowlogick, line 25: This change is unrelated to this PR, could you raise another PR for this change and revert it here?

auth_server/authn/gitea_auth.go
```diff
@@ -0,0 +1,121 @@
...
+	} else {
+		return "https://git2.groschupp.org/api"
+	}
+}
+
+func (gha *GiteaAuth) getGiteaWebUri() string {
+	if gha.config.GiteaWebUri != "" {
+		return gha.config.GiteaWebUri
+	} else {
+		return "https://git2.groschupp.org"
```
techknowlogick, line 64: same as above

techknowlogick (changes requested):
Thanks for this PR :)

I've added a few notes, in addition could you also add some documentation about using this new provider.

techknowlogick:
@cgroschupp Hey, just checking if there is anything I can do to help out with this PR? I'm one of the project leads of Gitea and so I have some insight into that codebase.

techknowlogick:
Gitea has a PR for supporting docker registry auth itself here: https://github.com/go-gitea/gitea/pull/16048

PR #307, opened 2021-04-28
Added OpenID Connect support

FBasels:
Hey there,

Within our project we needed OpenID Connect support, so here it is. We currently run docker_auth inside Kubernetes and use Keycloak as OpenID Connect provider. It works fine and should also work with other OIDC provider. However it has to be mentioned that I have not tested it with others than Keycloak.


auth_server/authn/oidc_auth.go
```diff
@@ -0,0 +1,352 @@
...
+		return
+	}
+
+	glog.V(2).Infof("New OIDC auth token for %s (Current time: %s, expiration time: %s)", prof.Email, time.Now().String(), tok.Expiry.String())
+
+	dbVal := &TokenDBValue{
+		TokenType:    tok.TokenType,
+		AccessToken:  tok.AccessToken,
+		RefreshToken: tok.RefreshToken,
+		ValidUntil:   tok.Expiry.Add(time.Duration(-30) * time.Second),
```
AlexGustafsson, line 210: Why is this negative?

FBasels: Yes, I have oriented myself on the implementation of Google auth. But I admit that it is not the ideal way to do it, since access tokens usually have a short lifetime (couple of minutes) and usually the lifetime of refresh tokens is longer. 
Maybe it would be better to use the lifetime of the refresh token minus x seconds/minutes or to enable to predefine a lifetime of the database tokens

PR #325, opened 2021-10-31
Remove httpdown

YC:
**Reference:**
https://stackoverflow.com/questions/39320025
https://pkg.go.dev/net/http#Server.Shutdown

### DIFF
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -135,9 +135,17 @@ func ServeOnce(c *server.Config, cf string) (*server.AuthServer, *http.Server) {
 		TLSConfig: tlsConfig,
 	}
 	go func() {
-		if err := hs.ListenAndServeTLS(c.Server.CertFile, c.Server.KeyFile); err != nil {
-			if err == http.ErrServerClosed {
-				return
+		if c.Server.CertFile == "" && c.Server.KeyFile == "" {
+			if err := hs.ListenAndServe(); err != nil {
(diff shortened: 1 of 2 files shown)


auth_server/main.go
```diff
@@ -129,22 +128,25 @@ func ServeOnce(c *server.Config, cf string, hd *httpdown.HTTP) (*server.AuthServ
...
 		Handler:   as,
 		TLSConfig: tlsConfig,
 	}
-
-	s, err := hd.ListenAndServe(hs)
-	if err != nil {
-		glog.Exitf("Failed to set up listener: %s", err)
-	}
+	go func() {
+		if err := hs.ListenAndServeTLS(c.Server.CertFile, c.Server.KeyFile); err != nil {
```
techknowlogick, line 138: I'm concerned this is a breaking change as it requires having a TLS cert/key (even if self-signed), as some users serve docker auth as plaintext and have a TLS reverse proxy in front of the service handling secure communication. Would you be able to add a conditional to run with and without certs?

Thanks for you PRs :)

YC: Thanks for the review. 
I didn't realise that my own configuration was cert/key with token and HTTP for server/reverse proxy.

techknowlogick (approved):

auth_server/main.go
techknowlogick: No worries, thanks for the quick turnaround :)

I'll checkout the other PR now.

PR #328, opened 2021-11-26
Support Unix Domain Sockets

YC:
Similar to Docker Registry, use combination of `addr` and `net` (supporting values `tcp` and `unix`). 
Requires #325

### DIFF
--- a/examples/reference.yml
+++ b/examples/reference.yml
@@ -12,7 +12,11 @@
 
 server:  # Server settings.
   # Address to listen on.
+  # Can be HOST:PORT for TCP or file path (e.g. /run/docker_auth.sock) for Unix socket.
   addr: ":5001"
+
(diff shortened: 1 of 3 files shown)


examples/reference.yml
```diff
@@ -11,9 +11,12 @@
 #      rootcertbundle: "/path/to/server.pem"
 
 server:  # Server settings.
-  # Address to listen on.
+  # Address to listen on. Can be HOST:PORT for TCP or file path for Unix socket.
```
techknowlogick, line 14: Could you add an example of a filepath here? (eg.`/run/docker_auth.sock` or whichever is the correct format)?

techknowlogick (commented):
one very tiny nitpick, but otherwise is good! Thanks again :)

techknowlogick (approved):

techknowlogick:
Thanks again for all of your PRs ❤️

PR #339, opened 2022-04-25
Add env variable support for configuration

ewassef:
https://github.com/cesanta/docker_auth/pull/225 updated to current HEAD

> This commit also makes possible to use json files as configuration
files via viper configuration library.

> This PR contains breaking changes on configuration to make
configuration overridable by the environment variables. To make it
support nested struct overrides struct tags are renamed to not
contain any underscore in their names. This is going to break old
configuration files.

> Because of viper's limitations of detecting new map keys from env
variables, a little procedure goes trough prefixed env variables and
override viper settings by the parsed value of them. If the
configuration file is in yaml format env variables can contain correct
yaml strings and they will be parsed correctly. This behaviour is same
for json files too.

> PS: viper doesn't use file format parsers for unmarshalling the structures.
It parses files as interface{} and the using mapstructure library to
unmarshall structs from maps. This is why yaml struct tags converted to
mapstructure ones.

The original author is [@Um](https://github.com/umurgdk)

Tagging @techknowlogick frm the other PR https://github.com/techknowlogick


techknowlogick:
Thanks for picking up this PR @ewassef 

As you can imaging this will take some time to review, especially due to the breaking nature, however I can commit to you (haha, no pun intended 😉) to review this.

ewassef:
Thanks for the follow-up.. I went ahead and fixed the tests and committed the change.

ewassef:
@techknowlogick would you mind letting the checks kick off so I can fix anything that might come up/

ewassef:
@techknowlogick any guidance?

ewassef:
@flaper87 @techknowlogick @karmi anyone?

techknowlogick:
I've just tried pushing an update to see if that fixed the build

ewassef:
@techknowlogick any help?

mstup:
Is there any chance that it could be completed?

PR #348, opened 2022-09-21
allow LDAP mapping values to be lowercased

cardoe:
Add the ability to lowercase values when creating label mappings from LDAP attributes. In the example the groups that the user is a member of could be mixed case which would make an ACL like:
  match: { account: "/.+/", name: "${label:groups}/*" }
Not possible. But with this change and the example applied it would be possible.


techknowlogick (approved):
Thanks so much for this contribution!

cardoe:
What next steps do I need to do to get this merged and released?

PR #356, opened 2023-02-27
Missing pod annotation and security constraint in helm chart

germaino:
Add option to add pod annotation and pod security constraint in helm charts

Change-Id: I1fc7d7d2ca60411439669921fb6a353bb001225a


techknowlogick (approved):
Thanks for this PR <3

PR #375, opened 2023-11-01
Add option to adjust bcrypt hashing cost

andsens:
The default bcrypt hashing cost can be quite problematic when serving many requests at once. Depending on token expiration times and other security considerations, lowering the cost can be a valid fix. This PR adds the `token_hash_cost` to all TokenDB configs (redis, GCS, LevelDB). Note that the level db config was a simple string previously, I had to change it to a struct with `{path: string, token_hash_cost: int}`.

Note that this PR is built on top of #374.


andsens:
Done!

Issue #379, opened 2023-12-28
Remove insecure cipher suites


stingshen:
[Golang tls package](https://github.com/golang/go/blob/release-branch.go1.21/src/crypto/tls/cipher_suites.go#L53) lists cipher suites excluding those with security issues. 

We may exclude those insecure cipher suites from [TLSCipherSuitesValues](https://github.com/cesanta/docker_auth/blob/main/auth_server/server/config.go#L94)?

techknowlogick:
Thanks for the suggestion, would you be able to send a PR to align with the updated golang tls package list?

stingshen:
> Thanks for the suggestion, would you be able to send a PR to align with the updated golang tls package list?

Sure, the PR is submitted

PR #380, opened 2023-12-29
Use secure cipher suites for tls by default

stingshen:
Fixes https://github.com/cesanta/docker_auth/issues/379

Tested with `docker run --rm -ti  drwetter/testssl.sh -e --std 10.3.68.72:5001`

image

### DIFF
--- a/auth_server/main.go
+++ b/auth_server/main.go
@@ -64,9 +64,7 @@ func ServeOnce(c *server.Config, cf string) (*server.AuthServer, *http.Server) {
 		glog.Exitf("Failed to create auth server: %s", err)
 	}
 
-	tlsConfig := &tls.Config{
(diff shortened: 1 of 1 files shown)


auth_server/main.go
```diff
@@ -64,9 +64,7 @@ func ServeOnce(c *server.Config, cf string) (*server.AuthServer, *http.Server) {
 		glog.Exitf("Failed to create auth server: %s", err)
 	}
 
-	tlsConfig := &tls.Config{
-		PreferServerCipherSuites: true,
```
stingshen, line 68: PreferServerCipherSuites is deprecated and ignored

techknowlogick (approved):
Thanks 🙏

PR #384, opened 2024-02-28
Add additional examples logins to results page

genebean:
Prior to this, there was not an easy copy/paste line for people not using Docker. This addes example lines that can be copy/pasted for those of us who use Podman / Podman Desktop and containerd's nerdctl.

For context, nerdctl is the default way of interacting with containers run within Lima (https://lima-vm.io) and is how I run x86 containers on an Apple Silicon Mac.


techknowlogick:
Thanks!

PR #385, opened 2024-02-29
Replicate #384 for GitLab and OIDC

genebean:
I missed that there were three results pages so this replicates #384 for the other two.


techknowlogick (approved):
Thanks!! :)

PR #387, opened 2024-03-18
Add support for redis sentinel config by using redis.UniversalClient which auto-switches based on config

andsens:
This is one of the nice code changes where we can add more capability with the tiniest diff:  
redis-go has a `UniversalClient` which automatically detects which redis client to use (`simple`, `cluster`, or `sentinel`) based on [this piece of code](https://github.com/redis/go-redis/blob/16ab0f2ac309166bfef5cde51742f84c61c1e73c/universal.go#L173-L180).
```
func NewUniversalClient(opts *UniversalOptions) UniversalClient {
	if opts.MasterName != "" {
		return NewFailoverClient(opts.failover())
	} else if len(opts.Addrs) > 1 {
		return NewClusterClient(opts.cluster())
	}
	return NewClient(opts.simple())
}
```

Meaning, this PR introduces support for redis sentinel (HA failover) and also obsoletes the special `redis_cluster_options` setting. I haven't removed it though (legacy support). At some point a deprecation warning might be in order though.


andsens:
Woops. `UniversalOptions` only accepts `Addrs`, not `Addr` like with the `SimpleOptions`. I'll update the PR

andsens:
Hmpf, I don't see a way of doing this cleanly while preserving legacy configs. I suggest breaking compatibility and simplifying the entire thing (*looks around*, *ducks*).

Issue #386, opened 2024-03-12
docker-registry has changed its JWT implementation, no longer supports libtrust key IDs


zfLQ2qx2:
It looks like docker-registry made a breaking change on Oct 19th, 2023: 

https://github.com/distribution/distribution/pull/4096

As a result the kid field of the JWT token is no longer valid - the code to process the the values produced by docker-libtrust has been removed from docker-registry completely. It is not clear if the kid format just needs to be updated or if there are additional fields needed. 

It looks like the last docker-registry commit that will work with the current docker-auth is [1d410148efe6d1b7fd56457507a9dd465b105ec4](https://github.com/milosgajdos/distribution/tree/1d410148efe6d1b7fd56457507a9dd465b105ec4)

mikecook:
Note that the breaking change is in v3 but not in v2 of registry. 
see https://github.com/distribution/distribution/releases

zhangyoufu:
https://github.com/distribution/distribution/pull/4349 has changed the verification order back to certificate first. However, docker_auth still does not support distribution v3 due to missing x5c header in JWT.

https://github.com/distribution/distribution/blob/21f3291612301af9c5c9e6a839504ea140844b05/registry/auth/token/token.go#L175-L180

https://github.com/distribution/distribution/blob/21f3291612301af9c5c9e6a839504ea140844b05/registry/auth/token/token.go#L197-L212

https://github.com/go-jose/go-jose/blob/db085150a549dadd76f4d2ce2ac3976951cc461e/shared.go#L203-L210

Consider that x5c contains whole X.509 certificate in base64 form, I would prefer to configure auth.token.jwks in distribution configuration rather than include x5c header in every JWT.

zhangyoufu:
After fixing a logic error in aforementioned PR, and configuring trusted JWKS, I can successfully authenticate to distribution v3 using docker_auth. There is no modification required on docker_auth side.

eduardgruy:
I figured out how to work around this. Basically you need to use the JWKS parameter from https://distribution.github.io/distribution/about/configuration/#token

Now how to get it:
1. Generate your certificate and key
2. Use this script to generate jwks.json

https://github.com/distribution/distribution/issues/4470

zfLQ2qx2:
@eduardgruy I can't wait to test this! I ended up writing a replacement for docker_auth just so I could start to transition to registry v3.x but I didn't get very far because with the RFC style kid it stops complaining about the key id, but just gives a permission denied message.

[...]

PR #401, opened 2025-02-13
Add support for registry v3 JWK thumbprint key ID format

evanebb:
Fixes https://github.com/cesanta/docker_auth/issues/386.
This is a fairly quick and dirty fix, feel free to burn this PR down :)

As mentioned in that issue, the v3 version of the registry no longer supports `libtrust` key IDs. There are multiple alternative options to choose from, but the simplest one to implement for this project is using the [JWK thumbprint](https://www.rfc-editor.org/rfc/rfc7638.html) of the public key as the key ID instead.

For every certificate present in the `rootcertbundle` passed to the registry, it'll add the public key to the trusted keys identified by the JWK thumbprint: https://github.com/distribution/distribution/blob/63d3892315c817c931b88779399a8e9142899a8e/registry/auth/token/accesscontroller.go#L346-L348
So, if you pass this JWK thumbprint in the key ID header in the token, the registry can select the proper signing key using this thumbprint.

This PR allows configuring this through a new directive in the configuration file, namely `token.disable_legacy_key_id`.
If set to true, it will pass the JWK thumbprint in the key ID header instead of the `libtrust` key ID. It defaults to false for now, to avoid accidental breakage when updating setups using the v2 registry.

### DIFF
--- a/auth_server/server/config.go
+++ b/auth_server/server/config.go
@@ -17,11 +17,17 @@
 package server
 
 import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/big"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


choopm:
Thanks @evanebb for providing this PR.

I can confirm this fixes the issues when upgrading from `registry:2` to `registry:3`
by using the new config option `token.disable_legacy_key_id: true`.

@cesanta do you want to take a look at this PR please?

---
Until this gets merged, one might use the image I built which includes this change:

```shell
docker pull ghcr.io/choopm/docker_auth
# or
docker pull ghcr.io/choopm/docker_auth:1.13.0-1-g914f526
```
https://github.com/users/choopm/packages/container/package/docker_auth

techknowlogick (approved):

PR #410, opened 2025-07-16
Add disable_legacy_key_id to helm chart

### LINKED ISSUE #409: disable_legacy_key_id missing from chart ConfigMap
I've been having an issue with registry v3 and the docker_auth Helm chart which I believe to be due to the fact that token.disable_legacy_key_id: true is required, per https://github.com/cesanta/docker_auth/pull/401, but this setting is missing from the ConfigMap: https://github.com/cesanta/docker_auth/blob/main/chart/docker-auth/templates/configmap.yaml

This should be a very small PR, I should be able to raise one later this week, but I'm mentioning this in case someone else comes across it in the meantime!

harrygodwin10:
Fixes #409
Follows on from #401 to add the changes for disable_legacy_key_id to the helm chart configmap

### DIFF
--- a/chart/docker-auth/Chart.yaml
+++ b/chart/docker-auth/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: "1.14.0"
 description: Docker Registry V2 authentication server
 name: docker-auth
-version: 1.14.0
+version: 1.14.1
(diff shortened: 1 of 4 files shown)


joshrivers:
I just came here to prep my PR with the same change. @harrygodwin10 did it better than I did. +1

